PR #26, opened 2022-09-11
feat!: add --exclude and --include for list

blacha:
--exclude takes priority if something is excluded it will no be included

this removes --filter in favor of --include/--exclude

### DIFF
--- a/src/commands/list/__test__/filter.test.ts
+++ b/src/commands/list/__test__/filter.test.ts
@@ -3,44 +3,44 @@ import o from 'ospec';
 import { asyncFilter } from '../list.js';
 
 o.spec('AsyncFilter', () => {
-  function makGenerator(list: string[]): () => AsyncGenerator<{ path: string }> {
+  function makeGenerator(list: string[]): () => AsyncGenerator<{ path: string }> {
     return async function* gen(): AsyncGenerator<{ path: string }> {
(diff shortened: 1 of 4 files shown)


src/commands/list/__test__/filter.test.ts
```diff
@@ -0,0 +1,47 @@
+import { fsa } from '@chunkd/fs';
+import o from 'ospec';
+import { asyncFilter } from '../list.js';
+
+o.spec('AsyncFilter', () => {
+  function makGenerator(list: string[]): () => AsyncGenerator<{ path: string }> {
```
paulfouquet, line 6: What means `mak`?

blacha: haha `make` Ill fix it

src/commands/list/list.ts
```diff
@@ -58,14 +58,15 @@ export const commandList = command({
 
 export async function* asyncFilter<T extends { path: string }>(
   source: AsyncGenerator<T>,
-  filter: string,
+  opts?: { include?: string; exclude?: string }, //filter: string,
```
paulfouquet, line 61: Is `//filter: string,` intentionally being left?

blacha: Nope!

paulfouquet (approved):

PR #32, opened 2022-09-14
feat: core STAC validator TDE-456

amfage:
TODO after this PR: Fix up logging after PR https://github.com/linz/argo-tasks/pull/33 goes through


package.json
```diff
@@ -18,10 +18,13 @@
     "@linzjs/style": "^3.11.0",
     "@types/node": "^18.7.11",
     "@types/ospec": "^4.0.4",
+    "ajv": "8.8.2",
+    "ajv-formats": "2.1.1",
```
paulfouquet, line 22: It is safer to fix the versions

paulfouquet (commented):
That looks good. Do not forget to log the execution time when you'll refactor using the new logger

MDavidson17 (approved):
:+1:

blacha (approved):

PR #60, opened 2022-10-03
feat: simple-stac from tiff

blacha:


src/commands/stac-create/stac.create.ts
```diff
@@ -0,0 +1,138 @@
...
+    registerCli(args);
+
+    const collectionId = ulid();
+    const queue = new ConcurrentQueue(args.concurrency);
+
+    const abort = false;
+    const items: StacItem[] = [];
+    for (const location of args.location) {
+      for await (const filePath of fsa.list(location)) {
+        if (!filePath.endsWith('.tiff')) continue;
```
paulfouquet, line 31: Can they be `.tif` sometimes?

src/commands/copy/copy.ts
```diff
@@ -8,26 +10,34 @@ import { config, registerCli, verbose } from '../common.js';
...
+  return JSON.parse(gunzipSync(Buffer.from(x, 'base64url')).toString());
+}
+
 export const commandCopy = command({
   name: 'copy',
   args: {
     config,
     verbose,
+    concurrency: option({ type: number, long: 'concurrency', defaultValue: () => 10 }),
     manifest: restPositionals({ type: string, displayName: 'location', description: 'Manifest of file to copy' }),
```
paulfouquet, line 24: I am not sure to understand what is this input? `manifest` is confusing to me here. Where does it come from? Can you explain?

blacha:
Not used and very out of date.

PR #80, opened 2022-10-18
feat: validate tiffs are named the same as their tileset

blacha:


src/commands/tileset-validate/tileset.validate.ts
```diff
@@ -0,0 +1,71 @@
...
+import { MapSheet } from '../../utils/mapsheet.js';
+import { config, registerCli, verbose } from '../common.js';
+
+function isTiff(x: string): boolean {
+  const search = x.toLowerCase();
+  return search.endsWith('.tiff') || search.endsWith('.tif');
+}
+
+export const commandTileSetValidate = command({
+  name: 'list',
```
amfage, line 16: `name` should be `'tileset-validate` ?

amfage (approved):

PR #128, opened 2022-11-18
fix: report validation success at end

amfage:


src/commands/stac-validate/stac.validate.ts
```diff
@@ -146,8 +146,10 @@ export const commandStacValidate = command({
     await queue.join();
 
     if (failures.length > 0) {
-      logger.error({ failures: failures.length }, 'StacValidation:DoneWithErrors');
+      logger.error({ failures: failures.length }, 'StacValidationReport:DoneWithErrors');
```
blacha, line 149: could you make these similar so you can search for both outcomes without having to make a complex query?

`StacValidation:Done:Failed` `StacVaidldation:Done:Ok`


or possibly a secondary key

`StacValidation:Done` with `failed=true`

blacha (approved):

PR #135, opened 2022-11-28
fix: error if tileset validation fails TDE-586

amfage:
Small change to log error if tileset validation fails, so that it is obvious in the Argo Workflow for TDE-586.

### DIFF
--- a/src/commands/tileset-validate/tileset.validate.ts
+++ b/src/commands/tileset-validate/tileset.validate.ts
@@ -57,7 +57,7 @@ export const commandTileSetValidate = command({
             // Failed on size of tile
 
             failed++;
-            logger.error({ fileName, got: { x: bounds[0], y: bounds[3] }, expected }, 'File:Invalid:Size');
+            logger.error({ fileName, got: { width: width, height: height }, expected }, 'File:Invalid:Size');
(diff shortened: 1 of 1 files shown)


src/commands/tileset-validate/tileset.validate.ts
```diff
@@ -52,12 +52,12 @@ export const commandTileSetValidate = command({
...
 
             failed++;
-            logger.warn({ fileName, got: { x: bounds[0], y: bounds[3] }, expected }, 'File:Invalid:Origin');
+            logger.error({ fileName, got: { x: bounds[0], y: bounds[3] }, expected }, 'File:Invalid:Origin');
           } else if (width !== expected?.width || height !== expected.height) {
             // Failed on size of tile
 
             failed++;
-            logger.warn({ fileName, got: { x: bounds[0], y: bounds[3] }, expected }, 'File:Invalid:Size');
+            logger.error({ fileName, got: { x: bounds[0], y: bounds[3] }, expected }, 'File:Invalid:Size');
```
paulfouquet, line 60: Should we log the size instead of the coordinates?

paulfouquet (approved):

PR #208, opened 2023-01-24
feat: --force-no-clobber copy option TDE-595

amfage:


src/commands/copy/copy.ts
```diff
@@ -71,8 +85,8 @@ export const commandCopy = command({
             manifest,
             start: i,
             size: chunkSize,
-            force: args.force,
-            noClobber: args.noClobber,
+            force: force,
```
blacha, line 88: you don't need the `: force` if the key and value are the same name.

I should add a lint rule to auto fix this.

blacha (approved):

PR #363, opened 2023-04-13
fix: stac-validate handle list input TDE-504

amfage:


src/commands/stac-validate/stac.validate.ts
```diff
@@ -265,3 +270,10 @@ export function getStacChildren(stacJson: st.StacItem | st.StacCollection | st.S
 export function normaliseHref(href: string, path: string): string {
   return new URL(href, path).href;
 }
+
+export function listLocation(loc: string[]): string[] {
+  if (loc[0].startsWith('[')) {
```
paulfouquet, line 275: Would be nice to have a comment explaining why doing this specific parsing

src/commands/stac-validate/stac.validate.ts
```diff
@@ -55,7 +55,14 @@ export const commandStacValidate = command({
     const validated = new Set<string>();
 
     const recursive = args.recursive;
-    const paths = args.location.map((c) => c.trim());
+
+    if (args.location[0] === undefined) {
+      logger.error('StacValidation:Error:NoLocationProvided');
+      process.exit(1);
+    }
+    console.log(args.location);
```
paulfouquet, line 63: Are they "debug" logging?

amfage: Ha I forgot I just added that to test the input.

MDavidson17 (approved):

PR #417, opened 2023-05-07
feat: synchronize stac TDE-725

paulfouquet:
add a file hash to the object metadata on AWS S3 to determine if file has been modified

Has been tested here https://github.com/linz/imagery-test/actions/runs/4909985617/jobs/8766726824

### DIFF
--- a/src/commands/stac-sync/stac.sync.ts
+++ b/src/commands/stac-sync/stac.sync.ts
@@ -1,0 +1,89 @@
+import { fsa } from '@chunkd/fs';
+import { command, positional, string, Type } from 'cmd-ts';
+import { logger } from '../../log.js';
+import { config, registerCli, verbose } from '../common.js';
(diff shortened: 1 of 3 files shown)


src/commands/stac-sync/stac.sync.ts
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Get the hash of the file found in the HashKey value.
+ *
+ * @param Bucket
+ * @param Key
+ * @returns the hash if found
+ */
+export async function getHash(path: URL): Promise<string | null> {
+  const objHead = await fsa.head(path.href);
```
blacha, line 87: would be great to also validate the file size, and overwrite if the hash or file size has changed, which will hopefully fix the issue if someone overwrote the file in the CLI or ui?

blacha (approved):

PR #431, opened 2023-05-09
docs: fix typo in comment

paulfouquet:


blacha (changes requested):
:lgtm: however I would leave `fix:` for changes that actually have an effect on code, maybe this is a `docs`?

paulfouquet:
> :lgtm: however I would leave `fix:` for changes that actually have an effect on code, maybe this is a `docs`?

Yeah, that's a `chore` maybe?

blacha (approved):

PR #446, opened 2023-05-14
feat(stac-catalog): pretty print the Catalog output file

paulfouquet:


src/commands/stac-catalog/stac.catalog.ts
```diff
@@ -69,8 +76,8 @@ export const commandStacCatalog = command({
     const templateLinkCount = catalog.links.length;
 
     catalog.links = await createLinks(args.path, catalog.links);
-
-    await fsa.write(args.output, JSON.stringify(catalog));
+    
+    await fsa.write(args.output, JSON.stringify(catalog, null, spaces));
```
blacha, line 80: We should just always pretty print it.

 There is a slight conflict between what prettier thinks is pretty and what node thinks is pretty

Simplest option is just to jam a `2` where spaces is.

MDavidson17 (approved):

PR #448, opened 2023-05-16
feat(tileindex-validate): detect when tiffs are not in EPSG:2193 and error

blacha:


src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -84,6 +98,7 @@ export function findDuplicates(tiffs: CogTiff[], scale: number): FileList[] {
     const uri = f.source.uri;
     const firstImage = f.images[0];
     if (firstImage == null) throw new Error(`Failed to parse tiff: ${f.source.uri}`);
+    if (firstImage.epsg !== 2193) throw new Error(`Invalid projection tiff: ${f.source.uri} EPSG:${firstImage.epsg}`);
```
paulfouquet, line 101: Would it be always `2193`?

paulfouquet (approved):

PR #461, opened 2023-05-22
fix: dev dependencies were not installed before running prettier

amfage:


src/commands/stac-github-import/stac.github.import.ts
```diff
@@ -81,7 +81,7 @@ export const commandStacGithubImport = command({
     await fsa.write(collectionPath, JSON.stringify(collection));
     logger.info({ repo: gitRepo }, 'npm:install');
 
-    execFileSync('npm', ['install'], { cwd: gitRepo });
+    execFileSync('npm', ['install', '--dev'], { cwd: gitRepo });
```
blacha, line 84: I see, because we set `NODE_ENV` to production dev deps wont be installed by default.

I think `npm` prefers you to use `npm install --include=dev`  instead of `--dev`

paulfouquet (approved):

PR #479, opened 2023-05-30
fix: check for missmatching paths in manifest TDE-764

MDavidson17:


src/commands/create-manifest/create-manifest.ts
```diff
@@ -90,11 +90,22 @@ export async function createManifest(
...
       current.push({ source: filePath, target });
     }
     outputCopy.push(current);
   }
 
   return outputCopy;
 }
+
+export function validatePaths(source: string, target: string): void {
+  // 31.05.2023: fixes TDE-763
```
blacha, line 103: I dont think we need this comment, it would be great to document the function

src/commands/create-manifest/create-manifest.ts
```diff
@@ -90,11 +90,24 @@ export async function createManifest(
...
+  // Throws error if the source and target paths are not:
+  // - both directories
+  // - both paths
+  if (source.endsWith('/') && target.endsWith('/')) {
+    return;
+  }
+  if (!source.endsWith('/') && !target.endsWith('/')) {
+    return;
+  }
+  throw new Error(`Missmatch Paths - source: ${source}, target: ${target}`);
```
amfage, line 112: NIT: `mismatched` or `path mismatch`

amfage (approved):

PR #633, opened 2023-09-05
fix(copy): copy and fix content-types TDE-859

blacha:
A lot of files we write into S3 use the default content-type of "application/octet-stream" this content-type is not very useful for the browser to know to render the file or download it.

A lot of these files are written by other systems and we cannot correct in all systems.

This adds 

```
--fix-content-type=true
--fix-content-type=false // default
```

Which will attempt to guess what the content type should be when copying files,


src/commands/copy/copy-worker.ts
```diff
@@ -9,6 +10,29 @@ import { CopyContract, CopyContractArgs, CopyStats } from './copy-rpc.js';
...
 
+export const FixableContentType = new Set(['binary/octet-stream', 'application/octet-stream']);
+
+/** If the file has been written with a unknown binary contentType attempt to fix it with common content types */
+export function fixFileMetadata(target: string, f: FileInfo): FileInfo {
+  // If the content is encoded we do not know what the content-type should be
+  if (f.contentEncoding != null) return f;
+  if (!FixableContentType.has(f.contentType ?? 'binary/octet-stream')) return f;
+
+  if (target.endsWith('.tiff') || target.endsWith('.tif')) {
```
paulfouquet, line 21: Could use the existing `isTiff()` that also test `.TIFF` or `.TIF`. Even if it would most likely not happen if the file has been standardised before.

paulfouquet (approved):

PR #655, opened 2023-09-14
feat: add optional --fix-content-type TDE-859

paulfouquet:
### Description
Add an option `--fix-content-type` in `format-json` (previously `pretty-print`) to force the content-type to be updated for json s3 objects.

### Intention
Some of the STAC files (`Item`) have a content-type set to `application/json` instead of `application/geo+json`.

### Checklist
If not applicable, provide explanation of why.

- [ ] Tests updated: no tests for aws objects
- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ LINZ uses [Argo workflows](https://argoproj.github.io/workflows/) for running bu
 - [create-manifest](#create-manifest)
 - [group](#group)
 - [list](#list)
-- [pretty-print](#pretty-print)
+- [format-json](#format-json)
 - [stac catalog](#stac-catalog)
 - [stac github-import](#stac-github-import)
 - [stac sync](#stac-sync)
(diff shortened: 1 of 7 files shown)


src/commands/format/format.json.ts
```diff
@@ -1,22 +1,22 @@
...
 function isJson(x: string): boolean {
   const search = x.toLowerCase();
   return search.endsWith('.json');
 }
 
-export const commandPrettyPrint = command({
-  name: 'pretty-print',
-  description: 'Pretty print JSON files',
+export const commandFormatJson = command({
+  name: 'format-jspon',
```
amfage, line 18:
```suggestion
  name: 'format-json',
```

amfage (dismissed):

README.md
```diff
@@ -68,22 +68,22 @@ list s3://linz-imagery/sample --include ".*.tiff$" --group 10 --group-size 100MB
...
-Format all JSON files within a directory using `prettier`.
+Format all JSON files within a directory using `prettier`. An optional flag `--fix-content-type` allows to set the content-type for s3 object with a value depending on the STAC object type.
 
 #### Example
 
 - Format and overwrite files:
 
 ```bash
-pretty-print source/
+format-json source/ [--fix-content-type]
```
blacha, line 80: How come the name change? prettier will format a lot more than json. eg markdown, yaml, javascript etc..

paulfouquet: Because it's looking for `json` files atm. I can make the function more generic and look at other file format but I feel it will need a bit more work, specially around the `--fix-content-type`

blacha: This is a breaking change to rename it, so Id prefer to leave it as is if we can?

README.md
```diff
@@ -70,14 +70,14 @@ list s3://linz-imagery/sample --include ".*.tiff$"  --exclude "BG33.tiff$" --out
...
-Format all JSON files within a directory using `prettier`.
+Format all JSON files within a directory using `prettier`. An optional flag `--fix-content-type` allows to set the content-type for s3 object with a value depending on the STAC object type.
 
 #### Example
 
 - Format and overwrite files:
 
 ```bash
-pretty-print source/
+pretty-print source/ [--fix-content-type]
```
l0b0, line 80: The flag should be before the directory.

l0b0 (approved):
Other than a minor nit this looks good.

PR #662, opened 2023-09-20
feat(basemaps): Get the region from collection.json and validate the source and target bucket.

Wentao-Kuang:
#### Description
- We should validate the bucket from target and source location, before creating pull request in basemaps-config repo.
- If this is an individual layer, we will try to get the region from the source url.

#### Intention
- Add a check to make sure the target bucket is `linz-basemaps'`
- Add check to make sure the source bucket is `linz-imagery` or `nz-imagery`
- get the region from source url

#### Checklist
- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/src/commands/basemaps-github/create-pr.ts
+++ b/src/commands/basemaps-github/create-pr.ts
@@ -11,8 +11,11 @@ import { Category, MakeCogGithub, parseCategory } from './make.cog.github.js';
 
 async function parseTargetInfo(target: string): Promise<{ name: string; title: string; epsg: EpsgCode }> {
   const splits = target.replace('s3://', '').split('/');
+  const bucket = splits[0];
   const epsg = Epsg.tryGet(Number(splits[1]));
   const name = splits[2];
+  if (bucket == null || bucket !== 'linz-basemaps')
+    throw new Error(`Invalid s3 bucket ${bucket} from the target ${target}.`);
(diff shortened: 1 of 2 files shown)


src/commands/basemaps-github/create-pr.ts
```diff
@@ -9,17 +9,56 @@ import { logger } from '../../log.js';
...
   if (epsg == null || name == null) throw new Error(`Invalid target ${target} to parse the epsg and imagery name.`);
   const collectionPath = fsa.join(target, 'collection.json');
   const collection = await fsa.readJson<StacCollection>(collectionPath);
+  if (collection == null) throw new Error(`Failed to get target collection json from ${collectionPath}.`);
   const title = collection.title;
   if (title == null) throw new Error(`Failed to get imagery title from collection.json.`);
 
-  return { name: standardizeLayerName(name), epsg: epsg.code, title };
+  //Validate the source location
+  const source = collection.links[collection.links.length - 1]?.href;
```
blacha, line 41: you cannot trust that the last link is the correct link.

```
collection.links.find(f => f.rel == '')
```

Wentao-Kuang: Very good point.

blacha (approved):

PR #687, opened 2023-10-04
fix(basemaps): Remove the trailing slash if exists in the target location.

Wentao-Kuang:
#### Motivation
The input of targes from cogify will include a trialling slash, we should remove it if exists to make the config file looks consistent.

#### Modification
Check the slash exists and remove.

#### Checklist
- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/src/commands/basemaps-github/create-pr.ts
+++ b/src/commands/basemaps-github/create-pr.ts
@@ -121,7 +121,7 @@ export const basemapsCreatePullRequest = command({
         const info = await parseTargetInfo(target, args.individual);
         layer.name = info.name;
         layer.title = info.title;
-        layer[info.epsg] = target;
+        layer[info.epsg] = target.endsWith('/') ? target.slice(0, -1) : target;
         region = info.region;
       }
       layer.category = category;


src/commands/basemaps-github/create-pr.ts
```diff
@@ -121,7 +121,7 @@ export const basemapsCreatePullRequest = command({
         const info = await parseTargetInfo(target, args.individual);
         layer.name = info.name;
         layer.title = info.title;
-        layer[info.epsg] = target;
+        layer[info.epsg] = target.endsWith('/') ? target.slice(0, -1) : target;
```
l0b0, line 124: This seems like the sort of thing which should fail validation rather than trying to correct bad input. Or is this an instance of correct input which nevertheless has to be transformed?

Wentao-Kuang: This is out of date, we are working on adding slash at the end of all the s3 paths. So this PR will close later.

l0b0: I'm confused - why keep this open if we know we'll close it later?

Wentao-Kuang:
This is out of date, closed

PR #727, opened 2023-11-02
feat(basemaps): Move create-mapsheet into argo task. BM-879

Wentao-Kuang:
#### Motivation

This cli is mostly using for argo workflow not basemaps, we are move it into argo-tasks.

#### Modification
- Relocate the create-mapsheet cli into argo tasks
- Update it using the ts-cmd

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/src/commands/basemaps-github/create-mapsheet.ts
+++ b/src/commands/basemaps-github/create-mapsheet.ts
@@ -72,7 +72,7 @@ export const basemapsCreateMapSheet = command({
         c[EpsgCode.Nztm2000] != null && (c.maxZoom == null || c.maxZoom > 19) && (c.minZoom == null || c.minZoom < 32),
     );
     const imageryIds = new Set<string>();
-    for (const layer of aerial.layers) {
+    for (const layer of layers) {
       if (exclude && exclude.test(layer.name)) continue;
       if (include && !include.test(layer.name)) continue;
       if (layer[EpsgCode.Nztm2000] != null) imageryIds.add(layer[EpsgCode.Nztm2000]);
--- a/src/commands/basemaps-github/create-mapsheet.ts
+++ b/src/commands/basemaps-github/create-mapsheet.ts
@@ -1,0 +1,106 @@
+import { ConfigBundled, ConfigProviderMemory } from '@basemaps/config';
+import { Bounds } from '@basemaps/geo';
+import { fsa } from '@chunkd/fs';
+import { command, option, optional, string } from 'cmd-ts';
+import * as fgb from 'flatgeobuf/lib/mjs/geojson.js';
+import { FeatureCollection, MultiPolygon } from 'geojson';
+
+import { CliInfo } from '../../cli.info.js';
+import { logger } from '../../log.js';
+import { registerCli, verbose } from '../common.js';
+
+interface Output {
+  sheetCode: string;
+  files: string[];
+}
+
+export const CommandCreateMapSheetArgs = {
+  verbose,
+  path: option({
+    type: string,
+    long: 'path',
+    description: 'Path of flatgeobuf, this can be both a local path or s3 location',
+  }),
+  config: option({
+    type: string,
+    long: 'config',
+    description: 'Path of basemaps config json, this can be both a local path or s3 location',
+  }),
+  output: option({
+    type: string,
+    long: 'output',
+    description: 'Output of the mapsheet file',
+  }),
+  include: option({
+    type: optional(string),
+    long: 'include',
+    description: 'Include the layers with the pattern in the layer name.',
+  }),
+  exclude: option({
+    type: optional(string),
+    long: 'exclude',
+    description: 'Exclude the layers with the pattern in the layer name.',
+  }),
+};
+
+export const basemapsCreateMapSheet = command({
+  name: 'bm-create-mapsheet',
+  version: CliInfo.version,
+  description: 'Create a cog mapsheet from provided flatgeobuf',
+  args: CommandCreateMapSheetArgs,
+  async handler(args) {
+    registerCli(this, args);
+    const path = args.path;
+    const config = args.config;
+    const outputPath = args.output;
+
+    const include = args.include ? new RegExp(args.include.toLowerCase(), 'i') : undefined;
+    const exclude = args.exclude ? new RegExp(args.exclude.toLowerCase(), 'i') : undefined;
+
+    logger.info({ path }, 'MapSheet:LoadFgb');
+    const buf = await fsa.read(path);
+    logger.info({ config }, 'MapSheet:LoadConfig');
+    const configJson = await fsa.readJson<ConfigBundled>(config);
+    const men = ConfigProviderMemory.fromJson(configJson);
+
+    const rest = fgb.deserialize(buf) as FeatureCollection;
+
+    const aerial = await men.TileSet.get('ts_aerial');
+    if (aerial == null) throw new Error('Invalid config file.');
+    const layers = aerial.layers.filter(
+      (c) => c[2193] != null && (c.maxZoom == null || c.maxZoom > 19) && (c.minZoom == null || c.minZoom < 32),
+    );
+    const imageryIds = new Set<string>();
+    for (const layer of layers) {
+      if (exclude && exclude.test(layer.name)) continue;
+      if (include && !include.test(layer.name)) continue;
+      if (layer[2193] != null) imageryIds.add(layer[2193]);
+    }
+    const imagery = await men.Imagery.getAll(imageryIds);
+
+    const output: Output[] = [];
+    logger.info({ path, config }, 'MapSheet:CreateMapSheet');
+    for (const feature of rest.features) {
+      if (feature.properties == null) continue;
+      const sheetCode = feature.properties['sheet_code_id'];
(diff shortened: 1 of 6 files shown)


src/commands/basemaps-github/create-mapsheet.ts
```diff
@@ -0,0 +1,106 @@
...
+  async handler(args) {
+    registerCli(this, args);
+    const path = args.path;
+    const config = args.config;
+    const outputPath = args.output;
+
+    const include = args.include ? new RegExp(args.include.toLowerCase(), 'i') : undefined;
+    const exclude = args.exclude ? new RegExp(args.exclude.toLowerCase(), 'i') : undefined;
+
+    logger.info({ path }, 'MapSheet:LoadFgb');
```
l0b0, line 60: The `logger.info` lines in this PR all look like debug-level, rather than info-level. Is this intentional?

src/commands/basemaps-github/create-mapsheet.ts
```diff
@@ -0,0 +1,106 @@
...
+    const buf = await fsa.read(path);
+    logger.info({ config }, 'MapSheet:LoadConfig');
+    const configJson = await fsa.readJson<ConfigBundled>(config);
+    const men = ConfigProviderMemory.fromJson(configJson);
```
l0b0, line 64: `men` seems like a really strange name for this. Do we have an idiomatic name?

src/commands/basemaps-github/create-mapsheet.ts
Wentao-Kuang: I think these are the main steps for the cli to indicate the processing steps. 
Debug level are mostly used for lower level logs like logging result inside a loop for debugging.

src/commands/basemaps-github/create-mapsheet.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    const rest = fgb.deserialize(buf) as FeatureCollection;
+
+    const aerial = await men.TileSet.get('ts_aerial');
+    if (aerial == null) throw new Error('Invalid config file.');
+    const layers = aerial.layers.filter(
+      (c) => c[2193] != null && (c.maxZoom == null || c.maxZoom > 19) && (c.minZoom == null || c.minZoom < 32),
```
l0b0, line 71: 2193 is used more than once. Is there a useful name for what this index represents that we can use for a `const`?

src/commands/basemaps-github/create-mapsheet.ts
```diff
@@ -0,0 +1,106 @@
...
+    );
+    const imageryIds = new Set<string>();
+    for (const layer of layers) {
+      if (exclude && exclude.test(layer.name)) continue;
+      if (include && !include.test(layer.name)) continue;
+      if (layer[2193] != null) imageryIds.add(layer[2193]);
+    }
+    const imagery = await men.Imagery.getAll(imageryIds);
+
+    const output: Output[] = [];
```
l0b0, line 81: An array of `Output` entries should probably have a plural name, such as `outputs`.

src/commands/basemaps-github/create-mapsheet.ts
l0b0: This is `men`, though, with an "n".

Wentao-Kuang: Haha, good point, this should be `mem` though.

src/commands/basemaps-github/create-mapsheet.ts
Wentao-Kuang: This is easier for us to know which projection for the layer inside config. If we want to change this, probably something like `EpsgCode.Nztm2000`.

The type for layer is a `Record<EpsgCode, string>`.

blacha: this loop is also run twice, it would be better to put this minZoom/maxZoom check into the loop on L77.

I think it would also be nice to use EpsgCode.NZTM2000 instead of 2193 maybe we should use this convention more?

Wentao-Kuang: I have updated the EpsgCode.

For the loop, I think it much easier to read the logic if we filter layers first, because filtered layer been used by both L77 and L93. So it much be better to filter it first and avoid the complex condition check in both loops.

src/commands/basemaps-github/create-mapsheet.ts
```diff
@@ -0,0 +1,109 @@
...
+  description: 'Create a cog mapsheet from provided flatgeobuf',
+  args: CommandCreateMapSheetArgs,
+  async handler(args) {
+    registerCli(this, args);
+    const path = args.path;
+    const config = args.config;
+    const outputPath = args.output;
+
+    const include = args.include ? new RegExp(args.include.toLowerCase(), 'i') : undefined;
+    const exclude = args.exclude ? new RegExp(args.exclude.toLowerCase(), 'i') : undefined;
```
l0b0, line 58: There shouldn't be any need for `toLowerCase` if you're already using case-insensitive search.

src/commands/basemaps-github/create-mapsheet.ts
blacha: As this is a copy / paste and used very infrequently I'm not too fused, but all these  loops could be rewritten into a single loop.  

All the imagery is loaded into memory https://github.com/linz/argo-tasks/pull/727/files#diff-bf30fd5dd0d7c140c3054a1e0fe58a85c5a4250fa860f81ce7fad23f0ecdc404R64

so we dont need to "getAll" on https://github.com/linz/argo-tasks/pull/727/files#diff-bf30fd5dd0d7c140c3054a1e0fe58a85c5a4250fa860f81ce7fad23f0ecdc404R82

We could drop it all and have a inner loop that looks something like

https://github.com/linz/argo-tasks/pull/727/files#diff-bf30fd5dd0d7c140c3054a1e0fe58a85c5a4250fa860f81ce7fad23f0ecdc404R93

```typescript
for (const layer of aerial.layers) {
 if (layer.minZoom == null || layer.minZoom < 32) continue;
 if (layer.maxZoom == null || layer.maxZoom > 19) continue; 
 
 const nztmImageryId = layer[EpsgCode.NZTM2000];
 if (nztmImageryId == null) continue;
 
 if (exclude && exclude.test(layer.name)) continue;
 if (include && !include.test(layer.name)) continue;
 
 const img = mem.objects.get(nztmImageryId)
 if (img == null) continue;
 


 // Do bounds check and add current files
}


```

Wentao-Kuang: I have added some unit test for this then refactoring this into simpler logic. I did a compare of the full output mapsheet.json after refactoring which getting the same output, and the unit test aren't broken neither.

l0b0: LGTM, although I'd love to see more use of random/"any" values in the test inputs so that it's obvious which parts are relevant to the assertion. What do you think, @blacha?

src/commands/basemaps-mapsheet/__test__/create-mapsheet.test.ts
```diff
@@ -0,0 +1,155 @@
+// import assert from 'node:assert';
```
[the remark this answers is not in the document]
Wentao-Kuang: Good pick, I forget to remove this.

src/commands/basemaps-mapsheet/__test__/create-mapsheet.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    ],
+  };
+  const mem = new ConfigProviderMemory();
+  mem.put(aerial);
+  mem.put(satellite);
+  mem.put(southland);
+
+  it('Should create the correct mapsheets', async () => {
+    const outputs = await createMapSheet(aerial, mem, rest, undefined, undefined);
+    console.log(outputs);
```
blacha, line 128:
```suggestion
```

src/commands/basemaps-mapsheet/__test__/create-mapsheet.test.ts
```diff
@@ -0,0 +1,154 @@
...
+          's3://linz-basemaps/2193/nz_satellite_2021-2022_10m_RGB/01G73E4AMSQ91TXQ3KC90NNPA0/6-28-36.tiff',
+          's3://linz-basemaps/2193/southland_rural_2005-2011_0-75m_RGBA/01F6P1S7DSP4N5DCBGCTZCMJ27/8-113-145.tiff',
+        ],
+      },
+    ]);
+  });
+
+  it('Should exclude the satellite', async () => {
+    const outputs = await createMapSheet(aerial, mem, rest, undefined, new RegExp('satellite', 'i'));
+    console.log(outputs);
```
blacha, line 143:
```suggestion
```

blacha (dismissed):

src/commands/basemaps-github/create-mapsheet.ts
blacha: While it would be nice to have a bunch more tests, this thing is run maybe once every other month, I'm a little hesitant in adding a bunch more work to this PR as its been sitting around a while now too.

PR #730, opened 2023-11-03
feat: lint imagery input target paths TDE-857

MDavidson17:
#### Motivation

in order to be sure that workflow input public taraget paths follow conventions by linting workflow inputs where possible.

#### Modification

Adding lint-paths step to argo-tasks which will be used in `pubish-odr` and can also be added to `publish-copy`

Currently just the publish paths are linted (linz-imagery & nz-imagery) - this was based on the ticket description but more can be added if required. 
The dataset section of the s3 path is currently not linted as this is too variable to check?

(this change will be used in a topo-workflow PR currently WIP)

#### Checklist

- [x] Tests updated
- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ LINZ uses [Argo workflows](https://argoproj.github.io/workflows/) for running bu
 - [lds-fetch-layer](#lds-fetch-layer)
 - [create-manifest](#create-manifest)
 - [group](#group)
+- [lint](#lint)
 - [list](#list)
 - [pretty-print](#pretty-print)
 - [stac catalog](#stac-catalog)
@@ -36,6 +37,16 @@ Multiple layers can be fetched at the same time, fetch `51002` and `51000`:
 
 ```bash
 lds-fetch-layer  --target ./output 51002 51000
+```
+
+### `lint`
+
+Lint imagery s3 target paths
+
+#### Example
+
+```bash
+lint-inputs s3://nz-imagery/auckland/auckland_2012_0.075m/rgb/2193/
 ```
 
 ### `list`
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ Multiple layers can be fetched at the same time, fetch `51002` and `51000`:
 lds-fetch-layer  --target ./output 51002 51000
 ```
 
-### `lint`
+### `lint-inputs`
 
 Lint imagery s3 target paths
 
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ LINZ uses [Argo workflows](https://argoproj.github.io/workflows/) for running bu
 - [lds-fetch-layer](#lds-fetch-layer)
 - [create-manifest](#create-manifest)
 - [group](#group)
-- [lint](#lint)
+- [lint-inputs](#lint-inputs)
 - [list](#list)
 - [pretty-print](#pretty-print)
 - [stac catalog](#stac-catalog)
--- a/src/commands/lint/__test__/lint.test.ts
+++ b/src/commands/lint/__test__/lint.test.ts
@@ -1,0 +1,32 @@
+import assert from 'node:assert';
+import { describe, it } from 'node:test';
+
+import { lintODRImageryPath } from '../lint.s3.paths.js';
+
+describe('lintODRImageryPaths', () => {
+  it('Should Fail - Incorrect Bucket Name', () => {
+    assert.throws(() => {
+      lintODRImageryPath('s3://n-imagery/auckland/auckland_2012_0.075m/rgb/2193/');
+    }, Error);
+  });
+  it('Should Fail - Incorrect Region', () => {
+    assert.throws(() => {
+      lintODRImageryPath('s3://nz-imagery/hawkesbay/hawkes-bay_2012_0.075m/rgb/2193/');
+    }, Error);
+  });
+  it('Should Fail - Incorrect product', () => {
+    assert.throws(() => {
+      lintODRImageryPath('s3://nz-imagery/hawkes-bay/hawkes-bay_2012_0.075m/rgbi/2193/');
+    }, Error);
+  });
+  it('Should Fail - Incorrect Crs', () => {
+    assert.throws(() => {
+      lintODRImageryPath('s3://nz-imagery/auckland/auckland_2012_0.075m/rgb/219/');
+    }, Error);
+  });
+  it('Should Pass', () => {
+    assert.throws(() => {
+      lintODRImageryPath('s3://nz-imagery/auckland/auckland_2012_0.075m/rgb/2193/');
+    }, Error);
+  });
+});
--- a/src/commands/lint/__test__/lint.test.ts
+++ b/src/commands/lint/__test__/lint.test.ts
@@ -9,6 +9,16 @@ describe('lintImageryPaths', () => {
       lintPath('s3://n-imagery/auckland/auckland_2012_0.075m/rgb/2193/');
     }, Error);
   });
+  it('Should Fail - Missing key (Product)', () => {
+    assert.throws(() => {
+      lintPath('s3://nz-imagery/auckland/auckland_2012_0.075m/2193/');
+    }, Error);
+  });
+  it('Should Fail - Extra args', () => {
+    assert.throws(() => {
+      lintPath('s3://nz-imagery/auckland/auckland_2012_0.075m/2193/extra-args/');
+    }, Error);
+  });
   it('Should Pass', () => {
     assert.ok(() => {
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -38,6 +39,16 @@ Multiple layers can be fetched at the same time, fetch `51002` and `51000`:
 lds-fetch-layer  --target ./output 51002 51000
 ```
 
+### `lint`
```
l0b0, line 42: Shouldn't this be in sync with the command name?

src/commands/lint/__test__/lint.test.ts
```diff
@@ -0,0 +1,45 @@
+import assert from 'node:assert';
+import { describe, it } from 'node:test';
+
+import { lintImageryPath, lintPath } from '../lint.s3.paths.js';
+
+describe('lintImageryPaths', () => {
+  it('Should Fail - Incorrect Bucket Name', () => {
+    assert.throws(() => {
+      lintPath('s3://n-imagery/auckland/auckland_2012_0.075m/rgb/2193/');
```
l0b0, line 9: Does AWS have a set of test buckets which will always be empty? It's not a stretch that someone will create a bucket called "n-imagery", which would mean this test would fail for the wrong reason.

src/commands/lint/lint.s3.paths.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  if (imageryBucketNames.includes(bucket)) {
+    lintImageryPath(pathArray);
+  } else {
+    throw new Error(`Bucket not bucket list: ${bucket}`);
+  }
+}
+
+export function lintImageryPath(pathArray: string[]): void {
```
l0b0, line 51: Could we split `pathArray` into `region`, `dataset`, etc? That way there would be no need to unpack the array.

src/commands/lint/lint.s3.paths.ts
```diff
@@ -0,0 +1,70 @@
...
+      throw new Error('No path Provided');
+    }
+    logger.info({ target: args.path }, 'LintInputs:Info');
+    lintPath(args.path);
+    logger.info({ duration: performance.now() - startTime }, 'LintInputs:Done');
+  },
+});
+
+export function lintPath(path: string): void {
+  const pathArray: string[] = path.replace('s3://', '').split('/');
```
l0b0, line 37: This can end up splitting the path into more than four components. You probably want to use [the `limit` argument](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split).

src/commands/lint/lint.s3.paths.ts
```diff
@@ -0,0 +1,70 @@
...
+  } else {
+    throw new Error(`Bucket not bucket list: ${bucket}`);
+  }
+}
+
+export function lintImageryPath(pathArray: string[]): void {
+  const region: string | undefined = pathArray[1];
+  const dataset: string | undefined = pathArray[2];
+  const product: string | undefined = pathArray[3];
+  const crs: string | undefined = pathArray[4];
```
l0b0, line 55: This doesn't check whether there are more than four `pathArray` entries. But see the other comments for ways around that.

src/commands/lint/lint.s3.paths.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  // lint target path
+  if (region == null || dataset == null || product == null || crs == null) {
+    throw new Error(`Missing key from path: s3://${pathArray.join('/')}`);
```
l0b0, line 59: It might be good to use custom error classes for all the validation errors:

- The tests could each easily check for the right error class, making them stricter.
- You could forward only the relevant inputs to the error class, and have the error class deal with formatting a message.

So in this line you might end up with `throw new MissingKeyFromPathError(pathArray);`.

src/commands/lint/lint.s3.paths.ts
```diff
@@ -0,0 +1,70 @@
...
+    registerCli(this, args);
+    const startTime = performance.now();
+
+    logger.info('LintInputs:Start');
+    if (!args.path) {
+      throw new Error('No path Provided');
+    }
+    logger.info({ target: args.path }, 'LintInputs:Info');
+    lintPath(args.path);
+    logger.info({ duration: performance.now() - startTime }, 'LintInputs:Done');
```
l0b0, line 32: Shouldn't runtime and the like be lower level than info? I think it's usually debug level.

src/commands/lint/__test__/lint.test.ts
blacha: This is specific to these buckets, we are trying to ensure that the imagery folder paths are consistent. while allowing for expansion later into per bucket folder paths.

src/commands/lint/lint.s3.paths.ts
blacha: the `info` is more that the command finished ok, it is just dumping more helpful info at the same time.

src/commands/lint/__test__/lint.test.ts
l0b0: Yep, just realised that was the case after looking at the error checking.

src/commands/lint/lint.s3.paths.ts
blacha: I would approach this differently.

you should use `new URL(path)` to validate that the object is a properly formatted url.

then from the url you can gain access to `hostname` and `pathname`

`hostname` is the bucket
`pathname` can then be split into its components, 

Oversplitting the pathname could (should?) be validated to so we dont make too many subfolders.

src/commands/lint/lint.s3.paths.ts
blacha: > So in this line you might end up with throw new MissingKeyFromPathError(pathArray);.

Subclassing in javascript is not super used mostly because the instance of checker is somewhat hit and miss, generally everything just throws a standard error with codes, eg a `fs.readFile('foo')` throws a `Error` with code `ENOENT`

l0b0: If we don't want to use subclassing then the tests should at least check the error message, to make sure it's catching the relevant error and not something else.

blacha: Yes what error is thrown needs to be checked.

Simplest way is to use a RegExp in the test

```typescript
assert.throws(() => new Error('Hello'), /^Error: Hello$/)
```

README.md
l0b0: The link above is now broken.

src/commands/lint/lint.s3.paths.ts
l0b0: Agreed, using `URL` to parse the URL is probably going to be much more reliable.

src/commands/lint/lint.s3.paths.ts
```diff
@@ -0,0 +1,87 @@
...
+  },
+});
+
+export function lintPath(path: string): void {
+  const [bucket, region, dataset, product, crs] = path.replace('s3://', '').split('/', 5);
+
+  if (bucket == null || region == null || dataset == null || product == null || crs == null) {
```
l0b0, line 60: [When there are not enough values to unpack (line 58), the values end up as `undefined` rather than `null`](https://www.typescriptlang.org/play?#code/FAYw9gdgzgLgBAbQGYEsBOsA0coFNwQAmAunALxwDkA7mGoZQHRQAOANijABSVUCUoSFDBtcjNmADmXPAUICgA), so we should probably `=== undefined` here.

src/commands/lint/lint.constants.ts
```diff
@@ -0,0 +1,24 @@
...
+  'canterbury',
+  'otago',
+  'southland',
+];
+
+export const imageryBucketNames = ['linz-imagery', 'nz-imagery'];
+
+export const imageryProducts = ['rgb'];
+
+export const imageryCrs = ['2193'];
```
l0b0, line 24: Same as https://github.com/linz/argo-tasks/pull/727#discussion_r1379494614, we should probably use `EpsgCode.NZTM2000` here. Are there constants for the other values as well?

MDavidson17: Thanks for sharing that, I hadn't seen it.
I think only the EpsgCode exists at the moment? but it would be cool to see more added.

l0b0 (approved):

src/commands/lint/lint.constants.ts
l0b0: Yeah, we should probably add any codes which are currently in use anywhere, and import them where necessary.

PR #734, opened 2023-11-08
feat(basemaps): To be able to ready basemaps config json file that been gunzipped.

Wentao-Kuang:
#### Motivation
The argo-tasks using latest version of `@chunkd/fs` which removed the logic to ready gz json files. However, create-mapsheet cli require to ready gz config file from s3.

#### Modification
Add the logic of reading gz file into JSON object.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/src/commands/basemaps-mapsheet/create-mapsheet.ts
+++ b/src/commands/basemaps-mapsheet/create-mapsheet.ts
@@ -4,17 +4,21 @@ import { fsa } from '@chunkd/fs';
 import { command, option, optional, string } from 'cmd-ts';
 import * as fgb from 'flatgeobuf/lib/mjs/geojson.js';
 import { FeatureCollection, MultiPolygon } from 'geojson';
+import { promisify } from 'util';
+import { gunzip } from 'zlib';
 
 import { CliInfo } from '../../cli.info.js';
 import { logger } from '../../log.js';
-import { registerCli, verbose } from '../common.js';
+import { config, registerCli, verbose } from '../common.js';
 
+const pGunzip = promisify(gunzip) as (data: Buffer) => Promise<Buffer>;
 interface Output {
   sheetCode: string;
   files: string[];
 }
 
 export const CommandCreateMapSheetArgs = {
+  config,
   verbose,
   path: option({
     type: string,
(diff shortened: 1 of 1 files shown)


src/commands/basemaps-mapsheet/create-mapsheet.ts
```diff
@@ -77,6 +81,19 @@ export const basemapsCreateMapSheet = command({
...
 });
 
+/**
+ * Read Basemaps bundled config file by a given path can be both local or s3 path. Also support for gunzip json file.
+ *
+ */
+async function readConfig(config: string): Promise<ConfigBundled> {
+  const data = await fsa.read(config);
+  if (config.endsWith('.gz')) {
+    return JSON.parse((await pGunzip(data)).toString()) as ConfigBundled;
```
l0b0, line 91: Is `promisify` + `await` a common pattern? I've not seen that one before. Would it be easier to do `return gunzip(data, (error, result) => JSON.parse(result))`?

l0b0 (commented):
Could you please clarify the title and description? Based on the code it seems the title could be something like "feat(basemaps): Handle gzipped config file transparently".

src/commands/basemaps-mapsheet/create-mapsheet.ts
Wentao-Kuang: @blacha Adding this feature into [@chunkd](https://github.com/blacha/chunkd/pull/1345/files), I think we will close this and just upgrade to use the new @chunkd later.

blacha: Javascript is a async language, so most things that have a `callback` don't actually return anything

```typescript 
const x = gunzip(data, (error, result) => JSON.parse(result))
console.log(typeof x, x) // "undefined", undefined
```

`promisfy` is a function that converts the old callback syntax into something you can return and use in the new `async/await` syntax

```typescript
const y = await pGunzip(data)
console.log(y) // data
```

@Wentao-Kuang We need this pull request still as upgrading to latest chunkd will be too much work for this story.

PR #779, opened 2023-11-29
fix: Run code under test TDE-968

l0b0:
#### Motivation

The original code just asserted that we were able to create an anonymous
function to call the function under test, and didn't actually run the
function.

#### Checklist

- [x] Tests updated
- [ ] Docs updated (N/A)
- [x] Issue linked in Title

### DIFF
--- a/src/commands/lint/__test__/lint.test.ts
+++ b/src/commands/lint/__test__/lint.test.ts
@@ -20,9 +20,7 @@ describe('lintPath', () => {
     }, Error('Too many keys in Path: /auckland/auckland_2012_0.075m/rgb/2193/extra-args/'));
   });
   it('Should Pass', () => {
-    assert.ok(() => {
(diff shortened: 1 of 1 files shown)


src/commands/lint/__test__/lint.test.ts
```diff
@@ -48,8 +46,6 @@ describe('lintImageryPaths', () => {
     }, Error('Trailing 0 in resolution of dataset name: auckland_2020_0.20m'));
   });
   it('Should Pass', () => {
```
blacha, line 48: While its not exactly part of this PR, test names should really be unique.

l0b0: Fixed in the [main PR](https://github.com/linz/argo-tasks/pull/780).

blacha (approved):

paulfouquet (approved):

PR #786, opened 2023-12-04
fix: specify bot email address in Pull Requests TDE-983

amfage:
#### Motivation

Pass through an email address associated with the repository that the PR is for.

#### Modification

Set the Git commit author email depending on whether the repo is for `basemaps-config`, `elevation`, or `imagery`.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [X] Issue linked in Title

### DIFF
--- a/src/commands/stac-github-import/stac.github.import.ts
+++ b/src/commands/stac-github-import/stac.github.import.ts
@@ -47,14 +47,14 @@ export const commandStacGithubImport = command({
     registerCli(this, args);
 
     const gh = new GithubApi(args.repoName);
-    const validRepos = ['linz/elevation', 'linz/imagery'];
-    let botEmail: string;
 
-    if (validRepos.includes(args.repoName)) {
-      botEmail = `${args.repoName.split('/')[1]}@linz.govt.nz`;
-    } else {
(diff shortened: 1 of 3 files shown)


src/commands/stac-github-import/stac.github.import.ts
```diff
@@ -47,6 +47,14 @@ export const commandStacGithubImport = command({
     registerCli(this, args);
 
     const gh = new GithubApi(args.repoName);
+    const validRepos = ['linz/elevation', 'linz/imagery'];
```
blacha, line 50: I think technically would would fine, however I think it is a little difficult to read, about something like this

```typescript
const BotEmails = {
 'linz/elevation': 'elevation@linz.govt.nz'
 'linz/imagery': 'imagery@linz.govt.nz'
}

const botEmail = BotEmails[args.repoName]
if (botEmail == null) throw new Error(...)
```

My primary concern is the array split, I think its considered a unchecked array access which will start erroring when we upgrade to the next typescript, even though we know there should always be 2 records in the split you still need to go and look at all the values in the array to be sure.

paulfouquet (approved):

blacha (approved):

PR #804, opened 2023-12-17
refactor: Remove dead code

l0b0:
#### Motivation

Cleans up miscellaneous issues found by JetBrains IDEA.

#### Modification

- Remove unreferenced example code
- Remove unused variables, types, classes
- Import type rather than relying on UMD global variable
- Wait for writes to finish

#### Checklist

None of these are applicable. This is just a cleanup.

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title


src/commands/basemaps-mapsheet/create-mapsheet.ts
```diff
@@ -95,7 +95,9 @@ export const basemapsCreateMapSheet = command({
     const outputs = await createMapSheet(aerial, mem, rest, include, exclude);
 
     logger.info({ outputPath }, 'MapSheet:WriteOutput');
-    fsa.write(outputPath, JSON.stringify(outputs, null, 2));
+    const outputWritePromise = fsa.write(outputPath, JSON.stringify(outputs, null, 2));
+
+    await Promise.all([featuresWritePromise, outputWritePromise]);
```
jfee-linz, line 100: recommend returning Promise.all, no need to await in the method unless you need it's response

l0b0: Nothing in the caller `await`s for these results though. The other commands all seem to `await` their results before their end.

jfee-linz: agreed.  The method is async, which implies it returns a promise.  The other awaited results seem to be used

src/commands/stac-validate/stac.validate.ts
```diff
@@ -69,14 +68,6 @@ export const commandStacValidate = command({
     const ajv = new Ajv.default({
       allErrors: true,
       strict: args.strict,
-      loadSchema: (uri: string): Promise<SchemaObject> => {
```
blacha, line 72: Why was this considered dead code?

l0b0: I couldn't find any reference to it. There's a separate function by the same name which is called, but this one never seems to be.

blacha: https://github.com/stac-utils/stac-node-validator/blob/master/index.js#L20C2-L20C12 implies it existed at one point?

l0b0: That does seem to reuse the function in the constructor. I still don't see anywhere it's called, although it could be within AJV itself. I'd expect our tests to suss that out, but maybe that code is untested.

l0b0: [This code](https://github.com/linz/argo-tasks/blob/0be61224024c2f4fc7263d9894766b88db9298e7/node_modules/ajv/dist/core.js#L162-L164) might indicate that it's being run.

l0b0: OIC, `src/commands/stac-validate/__test__/stac.validate.test.ts` doesn't test the validation functionality at all.

blacha: you can never assume that something doesn't run just because tests don't cover it... especially when its inside of a third party library!

l0b0: Of course! I meant that I expected the tests to cover it, so it would be detected if it was called. But I'll just revert this rather than write another test.

blacha (approved):

MDavidson17 (approved):

PR #810, opened 2023-12-21
fix: add 1:50k to valid grid sizes TDE-1010 TDE-1014 TDE-1015

amfage:
**Will revisit in the New Year - should probably revisit the validation approach**

#### Motivation

Allow the use of 50k scale imagery.

Note: it would be good to improve the tests and also take a look at how [tileset-validation logs errors](https://toitutewhenua.atlassian.net/browse/TDE-1013). I think these should be in a separate PR.

#### Modification

This is a simple fix to allow the use of 50k imagery. It also fixes a [bug](https://toitutewhenua.atlassian.net/browse/TDE-1015) where the bottom row of map sheets were not considered valid.

#### Checklist

- [x] Tests updated
- [ ] Docs updated
- [X] Issue linked in Title


src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -14,7 +14,7 @@ import { CommandListArgs } from '../list/list.js';
 
 const SHEET_MIN_X = MapSheet.origin.x + 4 * MapSheet.width; // The minimum x coordinate of a valid sheet / tile
```
blacha, line 15: I am not sure why we even have these, there is a MapSheet object with more precise  min and max values, which also have included unit tests.

We should remove what ever is referencing these.

src/utils/mapsheet.ts
```diff
@@ -89,7 +89,7 @@ export const MapSheet = {
   gridSizeMax: 50000,
   roundCorrection: 0.01,
   /** Allowed grid sizes, these should exist in the LINZ Data service (meters) */
-  gridSizes: [10_000, 5_000, 2_000, 1_000, 500],
```
blacha, line 92: When fixing bugs it would be nice to add a unit test.

amfage: It wasn't clear what the best way to do this was, I think the tests might need to be reviewed.

blacha: while this might hide the error, I don't think this fixes the bug, the tiles will not be named correctly when using 1:50k.

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -14,7 +14,7 @@ import { CommandListArgs } from '../list/list.js';
...
 const SHEET_MAX_X = MapSheet.origin.x + 46 * MapSheet.width; // The maximum x coordinate of a valid sheet / tile
-const SHEET_MIN_Y = MapSheet.origin.y - 41 * MapSheet.height; // The minimum y coordinate of a valid sheet / tile
+const SHEET_MIN_Y = MapSheet.origin.y - 42 * MapSheet.height; // The minimum y coordinate of a valid sheet / tile
```
l0b0, line 17: Now that's a magic number if ever I saw one. Did a test failure necessitate this? This should probably be explained in more detail, or be refactored out.

src/utils/mapsheet.ts
amfage: like this? https://toitutewhenua.atlassian.net/browse/TDE-1014

src/commands/tileindex-validate/tileindex.validate.ts
amfage: We could consult with the Land team and see what their requirements are? Perhaps we need to step back and look at the whole of tileset validation as we have another ticket regarding logging https://toitutewhenua.atlassian.net/browse/TDE-1013

src/commands/tileindex-validate/tileindex.validate.ts
paulfouquet: That was my mistake counting tiles I guess. This is my original draft when I worked on this:
image

amfage:
Superseded.

PR #827, opened 2024-01-18
feat: Slugify strings TDE-1019

l0b0:
#### Motivation

Allow us to easily slugify arbitrary strings for use in URLs.

#### Modification

This assumes that each part of a path will be slugified separately. This is less convenient than slugifying the entire path in one go, but ensures that no path segment contains a slash.

#### Checklist

- [x] Tests updated
- [x] Docs updated (comments only)
- [x] Issue linked in Title

### DIFF
--- a/src/utils/__test__/slugify.test.ts
+++ b/src/utils/__test__/slugify.test.ts
@@ -36,12 +36,12 @@ describe('slugify', () => {
   it('should treat any unhandled characters as an error', () => {
     assert.throws(
       () => {
-        slugify('“a\\b//c;\n”');
+        slugify('“a\\b//c—;\n”');
       },
       {
         name: 'UnhandledCharactersError',
-        message: 'Unhandled characters: "\\n", "/", ";", "\\", "“", "”"',
-        characters: ['\n', '/', ';', '\\', '“', '”'],
+        message: 'Unhandled characters: "\\n", "/", ";", "\\", "—", "“", "”"',
+        characters: ['\n', '/', ';', '\\', '—', '“', '”'],
       },
     );
   });
--- a/src/utils/__test__/slugify.test.ts
+++ b/src/utils/__test__/slugify.test.ts
@@ -1,11 +1,11 @@
 import assert from 'node:assert';
 import { describe, it } from 'node:test';
 
-import { outputAlphabet, slugify } from '../slugify.js';
+import { slugify } from '../slugify.js';
 
 describe('slugify', () => {
   it('should pass through output alphabet unchanged', () => {
-    assert.equal(slugify(outputAlphabet), outputAlphabet);
+    assert.equal(slugify('abcdefghijklmnopqrstuvwxyz0123456789_.-'), 'abcdefghijklmnopqrstuvwxyz0123456789_.-');
   });
   it('should lowercase uppercase ASCII characters', () => {
     assert.equal(slugify('ABCDEFGHIJKLMNOPQRSTUVWXYZ'), 'abcdefghijklmnopqrstuvwxyz');
--- a/src/utils/__test__/slugify.test.ts
+++ b/src/utils/__test__/slugify.test.ts
@@ -1,0 +1,12 @@
+import assert from 'node:assert';
+import { describe, it } from 'node:test';
+
+import { slugify } from '../slugify.js';
+
+const outputAlphabet = 'abcdefghijklmnopqrstuvwxyz0123456789_.-';
+
+describe('slugify', () => {
+  it('should pass through output alphabet unchanged', () => {
+    assert.equal(slugify(outputAlphabet), outputAlphabet);
+  });
+});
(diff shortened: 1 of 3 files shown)


src/utils/__test__/slugify.test.ts
```diff
@@ -0,0 +1,34 @@
...
+    assert.equal(slugify('abcdefghijklmnopqrstuvwxyz0123456789_.-'), 'abcdefghijklmnopqrstuvwxyz0123456789_.-');
+  });
+  it('should lowercase uppercase ASCII characters', () => {
+    assert.equal(slugify('ABCDEFGHIJKLMNOPQRSTUVWXYZ'), 'abcdefghijklmnopqrstuvwxyz');
+  });
+  it('should replace spaces with hyphens', () => {
+    assert.equal(slugify('Upper North Island'), 'upper-north-island');
+  });
+  it('should remove diacritics', () => {
+    assert.equal(slugify('äéìôūÄÉÌÔŪ'), 'aeiouaeiou');
```
blacha, line 17: I think this is missing some?

The NZGB placename database has these: 
```
áéóöøüĀāēīŌōŪū
```

src/utils/slugify.ts
```diff
@@ -0,0 +1,36 @@
...
+  /*
+   Normalization form decomposition (NFD) splits characters like into their
+   [combining diacritical mark](https://www.unicode.org/charts/PDF/U0300.pdf) and the character which is being modified
+   by the diacritic. This way we can remove the macron from "ā", the accent from "é", and the like.
+   */
+  const combiningDiacriticalMarks = /[\u0300-\u036F]/g;
+  return input.normalize('NFD').replaceAll(combiningDiacriticalMarks, '');
+}
+
+class UnhandledCharactersError extends Error {
```
blacha, line 25: We dont generally subclass errors due to how the instanceof checking for them can often fail

src/utils/slugify.ts
```diff
@@ -0,0 +1,36 @@
+export function slugify(input: string): string {
```
blacha, line 1: could you jsdoc the function, eg https://github.com/linz/argo-tasks/blob/master/src/utils/geotiff.ts#L6

src/utils/slugify.ts
```diff
@@ -0,0 +1,36 @@
...
+    const sortedUniqueCharacters = Array.from(new Set(unhandledCharacters)).sort();
+    throw new UnhandledCharactersError(sortedUniqueCharacters);
+  }
+
+  return result;
+}
+
+function removeDiacritics(input: string): string {
+  /*
+   Normalization form decomposition (NFD) splits characters like into their
```
blacha, line 17: NIT: block comments generally have a * per line and will sometimes cause formatters to give weird outputs without them

src/utils/slugify.ts
blacha: Generally you just hope the "name"/"message" contains enough information to give the cause.

For more complex errors its a good idea to wrap them with additional information

```
try {

} catch(e) {
 throw new Error(`Failed to parse: ${url.href}`, { cause: e })
}
```

Trying to determine error types at runtime is very hit and miss as `catch(e: unknown)` since you are able to throw anything, `throw 1` is a perfectly valid throw

```
> try {throw 1} catch(e) { console.log(typeof e) }
number
```


The MDN docs for Error have a example with structured error causes, which would be a better way of handling this.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#providing_structured_data_as_the_error_cause

l0b0: `tsc` doesn't like that:

```
❯ npm run build && npm run test && npm run lint -- --fix

> @linzjs/argo-tasks@2.18.2 build
> tsc

src/utils/slugify.ts:15:80 - error TS2554: Expected 0-1 arguments, but got 2.

15     throw new Error(`Unhandled characters: ${formattedCharacters.join(', ')}`, {
                                                                                  ~
16       cause: {characters: sortedUniqueCharacters}
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
17     });
   ~~~~~


Found 1 error in src/utils/slugify.ts:15
```

blacha: ahh argo-tasks is currently defaulting back to node16 typings, we need to add

```json
    "lib": ["ES2022"],
    "target": "ES2022",
```

to the `tsconfig.json#compilerOptions` which will switch it to node18 types, when we move to node20 it should change to ES2023

l0b0: What else do I need to do? Adding those doesn't change the `tsc` result.

blacha: Just checked out your branch updated tsconfig to

```typescript
{
  "extends": "@linzjs/style/tsconfig.base.json",
  "compilerOptions": {
    "lib": ["ES2022"],
    "target": "ES2022",
    "outDir": "build"
  }
}
```

`tsc` works then switching back to 2021 `tsc` dies when I add a `throw new Error("foo", { cause: "bar" })`

l0b0: Didn't see the JSONPath in your reference; you're right.

blacha: This should have been changed when we migrated from node16 to node18.

l0b0: Fixed.

amfage (dismissed):

amfage (approved):

PR #829, opened 2024-01-19
fix: handle 1:50k imagery TDE-1014 TDE-1015

amfage:
#### Motivation

Allow the use of 1:50k scale imagery.

Note: there will be a separate pull request for [tileset-validation logs errors](https://toitutewhenua.atlassian.net/browse/TDE-1013).
#### Modification

Handle 1:50000 grid size, and generate the [correct name for this imagery](https://toitutewhenua.atlassian.net/browse/TDE-1014) (map sheet name only). It also fixes a [bug where the bottom row of map sheets were not considered valid](https://toitutewhenua.atlassian.net/browse/TDE-1015).

#### Checklist

- [x] Tests updated
- [x] Docs updated (comments only)
- [X] Issue linked in Title

### DIFF
--- a/src/commands/create-manifest/__test__/create-manifest.test.ts
+++ b/src/commands/create-manifest/__test__/create-manifest.test.ts
@@ -87,12 +87,12 @@ describe('createManifest', () => {
     ]);
   });
   describe('validatePaths', () => {
-    it('Should throw error for Missmatch Paths', () => {
+    it('Should throw error for Mismatched Paths', () => {
       assert.throws(() => {
         validatePaths('memory://source/', 'memory://target/sub/test.tiff');
       }, Error);
     });
-    it('Should also throw error for Missmatch Paths', () => {
+    it('Should also throw error for Mismatched Paths', () => {
       assert.throws(() => {
         validatePaths('memory://source/test.tiff', 'memory://target/sub/');
       }, Error);
--- a/src/commands/tileindex-validate/__test__/tileindex.validate.test.ts
+++ b/src/commands/tileindex-validate/__test__/tileindex.validate.test.ts
@@ -36,12 +36,16 @@ describe('getTileName', () => {
     assert.equal(convertTileName('CH11_1000_0501', 5000), 'CH11_5000_0101');
     assert.equal(convertTileName('CH11_1000_0505', 5000), 'CH11_5000_0101');
   });
-
   it('should get correct parent tile 1:10k', () => {
     assert.equal(convertTileName('CH11_1000_0101', 10000), 'CH11_10000_0101');
     assert.equal(convertTileName('CH11_1000_0110', 10000), 'CH11_10000_0101');
     assert.equal(convertTileName('CH11_1000_1010', 10000), 'CH11_10000_0101');
     assert.equal(convertTileName('CH11_1000_1001', 10000), 'CH11_10000_0101');
+  });
+  it('should get correct parent tile 1:50k', () => {
+    assert.equal(convertTileName('AT24_50000_0101', 50000), 'AT24_50000_0101');
+    assert.equal(convertTileName('AT25_50000_0101', 50000), 'AT25_50000_0101');
+    assert.equal(convertTileName('CK08_50000_0101', 50000), 'CK08_50000_0101');
   });
 });
 describe('tiffLocation', () => {
--- a/src/commands/tileindex-validate/tileindex.validate.ts
+++ b/src/commands/tileindex-validate/tileindex.validate.ts
@@ -341,37 +341,37 @@ export function validateTiffAlignment(tiff: TiffLocation, allowedError = 0.015):
   return true;
 }
 
-export function getTileName(centerX: number, centerY: number, grid_size: number): string {
-  if (!MapSheet.gridSizes.includes(grid_size)) {
+export function getTileName(x: number, y: number, gridSize: number): string {
+  if (!MapSheet.gridSizes.includes(gridSize)) {
     throw new Error(`The scale has to be one of the following values: ${MapSheet.gridSizes}`);
   }
 
-  const scale = Math.floor(MapSheet.gridSizeMax / grid_size);
-  const tile_width = Math.floor(MapSheet.width / scale);
-  const tile_height = Math.floor(MapSheet.height / scale);
-  let nb_digits = 2;
-  if (grid_size === 500) {
-    nb_digits = 3;
+  const scale = Math.floor(MapSheet.gridSizeMax / gridSize);
+  const tileWidth = Math.floor(MapSheet.width / scale);
+  const tileHeight = Math.floor(MapSheet.height / scale);
+  let nbDigits = 2;
+  if (gridSize === 500) {
+    nbDigits = 3;
   }
 
-  if (!(SHEET_MIN_X <= centerX && centerX <= SHEET_MAX_X)) {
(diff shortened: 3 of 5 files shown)


src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -14,7 +14,7 @@ import { CommandListArgs } from '../list/list.js';
 
 const SHEET_MIN_X = MapSheet.origin.x + 4 * MapSheet.width; // The minimum x coordinate of a valid sheet / tile
```
blacha, line 15: this logic needs to be scrapped, we need to allow tiles outside the "valid" range of topo50 mapsheets.

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -341,7 +341,7 @@ export function validateTiffAlignment(tiff: TiffLocation, allowedError = 0.015):
   return true;
 }
 
-export function getTileName(originX: number, originY: number, grid_size: number): string {
+export function getTileName(centerX: number, centerY: number, grid_size: number): string {
```
blacha, line 344: This was correct with originX, originY, this function does not need a center point, if anything it should be just `x` and `y` as it can be any point within the tile.

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -354,20 +354,20 @@ export function getTileName(originX: number, originY: number, grid_size: number)
...
-  if (!(SHEET_MIN_Y <= originY && originY <= SHEET_MAX_Y)) {
-    throw new Error(`y must be between ${SHEET_MIN_Y} and ${SHEET_MAX_Y}, was ${originY}`);
+  if (!(SHEET_MIN_Y <= centerY && centerY <= SHEET_MAX_Y)) {
+    throw new Error(`y must be between ${SHEET_MIN_Y} and ${SHEET_MAX_Y}, was ${centerY}`);
   }
 
   // Do some maths
-  const offset_x = Math.round(Math.floor((originX - MapSheet.origin.x) / MapSheet.width));
-  const offset_y = Math.round(Math.floor((MapSheet.origin.y - originY) / MapSheet.height));
+  const offset_x = Math.round(Math.floor((centerX - MapSheet.origin.x) / MapSheet.width));
```
blacha, line 365: if your refactoring here, these should all be `camelCase`

src/commands/tileindex-validate/__test__/tileindex.validate.test.ts
```diff
@@ -36,13 +36,17 @@ describe('getTileName', () => {
...
     assert.equal(convertTileName('CH11_1000_0505', 5000), 'CH11_5000_0101');
   });
-
   it('should get correct parent tile 1:10k', () => {
     assert.equal(convertTileName('CH11_1000_0101', 10000), 'CH11_10000_0101');
     assert.equal(convertTileName('CH11_1000_0110', 10000), 'CH11_10000_0101');
     assert.equal(convertTileName('CH11_1000_1010', 10000), 'CH11_10000_0101');
     assert.equal(convertTileName('CH11_1000_1001', 10000), 'CH11_10000_0101');
   });
+  it('should get correct parent tile 1:50k', () => {
```
blacha, line 45: 1:50k tile should be just the mapsheet name. `CK08.tiff`

paulfouquet:
Why not only linking to TDE-1010? 
It does link this PR to all the sub-tasks of TDE-1010 as you've linked it to TDE-1010, which include TDE-1014 and TDE-1015 but also TDE-1013. 
Other option will be to link to TDE-1014 and TDE-1015 if you don't mind not linking to TDE-1010.

src/commands/tileindex-validate/tileindex.validate.ts
amfage: I thought in the meeting we decided to do this later when the need arises. I double-checked with Rebecca yesterday and she would like to keep the logic in for checking imagery is a valid topo50 mapsheet.

src/commands/tileindex-validate/__test__/tileindex.validate.test.ts
amfage: It does generate the correct name from the map sheet further along in the process.

paulfouquet:
> Why not only linking to TDE-1010? It does link this PR to all the sub-tasks of TDE-1010 as you've linked it to TDE-1010, which include TDE-1014 and TDE-1015 but also TDE-1013. Other option will be to link to TDE-1014 and TDE-1015 if you don't mind not linking to TDE-1010.

Weirdly it still link this PR to TDE-1013 that is not mentioned in the title :thinking: . Is that because it has been linked before?

src/commands/tileindex-validate/tileindex.validate.ts
paulfouquet: I don't know how hard it would be to raise an alert if it's not a valid topo50 mapsheet but still allow it?

src/commands/tileindex-validate/__test__/tileindex.validate.test.ts
blacha: I dont think we should be generating incorrect names anywhere, this could really bite us in the future if we use this function for naming...

src/commands/create-manifest/__test__/create-manifest.test.ts
```diff
@@ -87,12 +87,12 @@ describe('createManifest', () => {
...
   });
   describe('validatePaths', () => {
-    it('Should throw error for Missmatch Paths', () => {
+    it('Should throw error for Mismatched Paths', () => {
       assert.throws(() => {
         validatePaths('memory://source/', 'memory://target/sub/test.tiff');
       }, Error);
     });
-    it('Should also throw error for Missmatch Paths', () => {
+    it('Should also throw error for Mismatched Paths', () => {
```
l0b0, line 95: Wait, what's the "also" here? If these tests are fundamentally different they should probably be named accordingly.

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -14,7 +14,7 @@ import { CommandListArgs } from '../list/list.js';
 
 const SHEET_MIN_X = MapSheet.origin.x + 4 * MapSheet.width; // The minimum x coordinate of a valid sheet / tile
 const SHEET_MAX_X = MapSheet.origin.x + 46 * MapSheet.width; // The maximum x coordinate of a valid sheet / tile
-const SHEET_MIN_Y = MapSheet.origin.y - 41 * MapSheet.height; // The minimum y coordinate of a valid sheet / tile
+const SHEET_MIN_Y = MapSheet.origin.y - 42 * MapSheet.height; // The minimum y coordinate of a valid sheet / tile
```
l0b0, line 17: As mentioned elsewhere, `42` and the other highly magical values should probably be pulled out into constants which make their semantics obvious.

src/utils/__test__/mapsheet.test.ts
```diff
@@ -19,6 +19,20 @@ describe('MapSheets', () => {
     });
   });
 
+  it('should generate 1:50k name', () => {
+    assert.deepEqual(MapSheet.extract('CK08_50000_0101.tiff'), {
+      mapSheet: 'CK08',
+      gridSize: 50000,
```
l0b0, line 25: Nit: ECMAScript 2021 supports [numeric separators](https://v8.dev/features/numeric-separators), so that big numbers can be written like `50_000` for clarity. Also, we use such separators elsewhere in this repo, so we should probably do the same here for consistency.

PR #845, opened 2024-01-31
fix: update target path naming hierarchy TDE-955

MDavidson17:
#### Motivation

The target path name formatting is inconsistent with the final decisions in the naming.md
It is now expected that the event name will be included in the `geographic_description` field as well. 

Sometimes the event name includes the location (`Top of the South Floods`) and sometimes it does not (`Cyclone Gabrielle`) Therefore it is difficult to determine whether to merge the `event_name` with the `region`/`geographic_description`. 

It is also preferable to keep the `event_name` consistent across all imagery so that datasets can be filtered by events for example the Cyclone Gabrielle imagery in Hawkes Bay, Gisborne & the North Island should all have `event_name=Cyclone Gabrielle`.

##### examples:
inputs: `region=Hawke's Bay` `gd= Hawke's Bay Cyclone Gabrielle` `event_name=Cyclone Gabrielle`
paths: `s3://nz-imagery/hawkes-bay/hawkes-bay-cyclone-gabrielle_2023_0.1m/`

inputs: `region=Nelson` `gd= Top of the South Floods` `event_name=Top of the South Floods`
paths: `s3://nz-imagery/nelson/top-of-the-south-floods_2023_0.1m/`

inputs: `region=Auckland`
paths: `s3://nz-imagery/auckland/auckland_2023_0.1m/`

inputs: `region=Wellington` `gd=Upper Hutt`
paths: `s3://nz-imagery/wellington/upper-hutt_2023_0.1m/`

#### Modification

name portion of the target path changed from:
```
if geographic_descr return geographic_descr-event?
else return region-event?
```
to
```
if geographic_descr return geographic_descr
return region
```

#### Checklist

- [x] Tests updated
~~- [ ] Docs updated~~ - n/a
- [x] Issue linked in Title

### DIFF
--- a/src/commands/path/__test__/generate.path.test.ts
+++ b/src/commands/path/__test__/generate.path.test.ts
@@ -40,10 +40,7 @@ describe('GeneratePathImagery', () => {
       gsd: 0.25,
       epsg: 2193,
     };
-    assert.equal(
-      generatePath(metadata),
-      's3://nz-imagery/hawkes-bay/hawkes-bay-north-island-weather-event_2023_0.25m/rgb/2193/',
-    );
+    assert.equal(generatePath(metadata), 's3://nz-imagery/hawkes-bay/north-island-weather-event_2023_0.25m/rgb/2193/');
   });
   it('Should match - no optional metadata', () => {
     const metadata: PathMetadata = {
@@ -94,9 +91,9 @@ describe('GeneratePathSatelliteImagery', () => {
     const metadata: PathMetadata = {
       targetBucketName: 'nz-imagery',
       category: 'satellite-imagery',
-      geographicDescription: 'North Island',
+      geographicDescription: undefined,
       region: 'new-zealand',
-      event: 'Cyclone Gabrielle',
+      event: 'North Island Cyclone Gabrielle',
       date: '2023',
       gsd: 0.5,
       epsg: 2193,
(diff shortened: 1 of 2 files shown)


src/commands/path/__test__/generate.path.test.ts
```diff
@@ -33,25 +32,20 @@ describe('GeneratePathImagery', () => {
     const metadata: PathMetadata = {
       targetBucketName: 'nz-imagery',
       category: 'rural-aerial-photos',
-      geographicDescription: undefined,
+      geographicDescription: 'North Island Weather Event',
```
l0b0, line 35: This is no longer just a geographic description. Should it be simply "description"?

MDavidson17: Here I'm trying to match the Stac Metadata field which is still `linz:geographic_description` whether the metadata should be changed now would be a question to check with @dwsilk 

(Most of the time it wont be an event just normal imagery?)

amfage (approved):

src/commands/path/__test__/generate.path.test.ts
dwsilk: The vast majority of the time it will be a geographic description, there are a handful of _different_ cases where it won't entirely be geographic in nature and in order to properly cover them we'd need to add a lot more complexity for very little benefit. We've split region and event names out into a standard list so that they are more useful metadata components, I think that's enough.

l0b0 (approved):

PR #849, opened 2024-02-01
refactor: Specify common types TDE-1030

l0b0:
#### Motivation

Strengthen type limitations to make it easier to understand where a "string" is actually a URL and where it's really a local path.

#### Checklist

Refactor; none of the checklist is applicable.


src/utils/types.ts
```diff
@@ -0,0 +1,8 @@
+declare const __brand: unique symbol;
```
l0b0, line 1: This needs more of an explanation - I'll add one.

src/commands/basemaps-mapsheet/create-mapsheet.ts
```diff
@@ -39,17 +41,17 @@ interface Output {
 export const CommandCreateMapSheetArgs = {
   verbose,
   path: option({
-    type: string,
+    type: PathStringOrUrlStringFromString,
```
blacha, line 44: This should be URL, anytime we are taking user input we should be validating that it does actually meet the standard of being a URL.

We use three common URL parsers in basemaps which could be replicated here:

https://github.com/linz/basemaps/blob/master/packages/cogify/src/cogify/parsers.ts

l0b0: Do we actually need `UrlFolder` or `UrlArrayJsonFile` anywhere in argo-tasks?

blacha: UrlFolder is very helpful to ensure that something is a folder and not a prefix

`s3://foo/bar` (prefix) vs `s3://foo/bar/` (folder) as when you join urls it has big implications as to where the file goes

for example adding `new URL("index.html", source)` to both of those would result in `s3://foo/index.html` and `s3://foo/bar/index.html`


UrlArrayJson is mostly  handled with the manifest files I think in this repo.

l0b0: I'll leave them out as long as we don't need them, though.

src/commands/basemaps-mapsheet/create-mapsheet.ts
```diff
@@ -22,9 +23,9 @@ export function isGzip(b: Buffer): boolean {
  *
  * If the file ends with .gz or is a GZIP like {@link isGzip} file it will automatically be decompressed.
  */
-async function readConfig(config: string): Promise<ConfigBundled> {
-  const obj = await fsa.read(config);
-  if (config.endsWith('.gz') || isGzip(obj)) {
+async function readConfig(config: URL): Promise<ConfigBundled> {
+  const obj = await fsa.read(config.href);
```
blacha, line 27: its not safe to use `.href` here. If its a file url it has to be converted with `path.fileURLToPath`.

does `fsa.read` not accept a URL? would imply `fsa` is out of date?

src/commands/create-manifest/create-manifest.ts
```diff
@@ -56,64 +65,60 @@ export const commandCreateManifest = command({
...
         if (actionLocation) {
-          const targetLocation = fsa.join(actionLocation, `actions/manifest-${targetHash}.json`);
+          const targetLocation = fsa.join(actionLocation.href, `actions/manifest-${targetHash}.json`);
           const targetAction: ActionCopy = { action: 'copy', parameters: { manifest: current } };
           await fsa.write(targetLocation, JSON.stringify(targetAction));
-          outputCopy.push(targetLocation);
+          outputCopy.push(new URL(targetLocation));
         } else {
-          outputCopy.push(gzipSync(outBuf).toString('base64url'));
+          outputCopy.push(new URL(gzipSync(outBuf).toString('base64url')));
```
blacha, line 73: `base64url` is never a URL

src/commands/create-manifest/create-manifest.ts
```diff
@@ -56,64 +65,60 @@ export const commandCreateManifest = command({
...
 
-    for (const filePath of chunk) {
-      const baseFile = args.flatten ? path.basename(filePath) : filePath.slice(source.length);
-      let target = targetPath;
+    for (const chunkUrl of chunk) {
+      const baseFile = args.flatten ? path.basename(chunkUrl.href) : chunkUrl.href.slice(sourceUrl.href.length);
+      let target = targetUrl;
       if (baseFile) {
-        target = fsa.joinAll(targetPath, transformFunc ? transformFunc(baseFile) : baseFile);
+        target = new URL(fsa.joinAll(targetUrl.href, transformFunc ? transformFunc(baseFile) : baseFile));
```
blacha, line 102: `joinAll` does a very similar job to `new URL(input, base)` in most use cases it should be a direct swap.

I believe the one caveat is `joinAll("foo.com/bar", "baz")` would make "foo.com/bar/baz" where as `new URL("baz", "foo.com/bar")` would make "foo.com/baz"

src/utils/action.storage.ts
```diff
@@ -4,12 +4,12 @@
  * Uses $ACTION_PATH if set to store the actions to a user defined location
  * This parses $ARGO_TEMPLATE looking for a `archiveLocation`
  */
-export function getActionLocation(): string | null {
-  if (process.env['ACTION_PATH']) return process.env['ACTION_PATH'];
+export function getActionLocation(): URL | null {
+  if (process.env['ACTION_PATH']) return new URL(process.env['ACTION_PATH']);
```
blacha, line 8: this will fail if "ACTION_PATH" is not a url, eg "/tmp" it is safer to use a URL parser for this.

src/commands/basemaps-mapsheet/create-mapsheet.ts
l0b0: > does `fsa.read` not accept a URL? would imply `fsa` is out of date?

[Yep](https://github.com/linz/argo-tasks/pull/817); that PR needs a bunch of fixes, though.

src/commands/create-manifest/create-manifest.ts
l0b0: OIC, [base64url encoding is used to create URL-safe strings from binary data](https://github.com/nodejs/node/issues/26512). Fixed.

src/commands/create-manifest/create-manifest.ts
l0b0: Indeed, `new URL('sub-path', 's3://bucket/path').href === "s3://bucket/sub-path" `. But I don't know what to do about that info. I would expect most of our base URLs to be non-root ones, in which case `new URL(input, base)` would not do what we want. Which means `joinAll` is the safer alternative. Or are you saying the base URLs will always end with a slash, so that both solutions do the same thing?

src/commands/copy/copy.ts
```diff
@@ -87,8 +75,12 @@ export const commandCopy = command({
     const chunks = [];
     const startTime = performance.now();
     for (const m of args.manifest) {
-      const data = await tryParse(m);
-      const manifest = CopyManifest.parse(data);
+      const json = await fsa.readJson<ActionCopy>(m);
+      if (json.action !== 'copy') throw new Error('Invalid action: ' + json.action + ' from:' + m);
```
paulfouquet, line 79:
```suggestion
      if (json.action !== 'copy') throw new Error('Invalid action: ' + json.action + ' from: ' + m);
```

src/utils/__test__/filter.test.ts
```diff
@@ -6,45 +6,60 @@ import { fsa } from '@chunkd/fs';
...
+      for (const path of paths) yield { url: new URL(`memory://${path}`) };
     };
   }
 
-  it('should include all', async () => {
-    const gen = makeGenerator(['a.tiff', 'b.tiff', 'c.tiff']);
-    const result = await fsa.toArray(asyncFilter(gen()));
-    assert.equal(result.length, 3);
-  });
+  // it('should include all', async () => {
```
paulfouquet, line 15: Is that meant to be removed?

l0b0: @blacha? This is commit 804b355b64496912bb9dabb62a2d33736a7c9023.

blacha: it looks to have been a mistake it should be uncommented.

src/commands/basemaps-mapsheet/create-mapsheet.ts
```diff
@@ -72,30 +73,30 @@ export const basemapsCreateMapSheet = command({
...
-    const buf = await fsa.read(path);
+    logger.info({ path: url }, 'MapSheet:LoadFgb');
+    const buf = await fsa.read(url);
     logger.info({ config }, 'MapSheet:LoadConfig');
     const configJson = await readConfig(config);
     const mem = ConfigProviderMemory.fromJson(configJson);
 
     const rest = fgb.deserialize(buf) as FeatureCollection;
-    const featuresWritePromise = fsa.write('features.json', JSON.stringify(rest));
+    const featuresWritePromise = fsa.write(new URL('file://features.json'), JSON.stringify(rest));
```
blacha, line 90: does `file://` write to the current folder?

src/commands/copy/copy-rpc.ts
```diff
@@ -6,7 +6,7 @@ export interface CopyContractArgs {
   /** Copy ID for tracing */
   id: string;
   /** List of files that need to be copied */
-  manifest: { source: string; target: string }[];
+  manifest: { source: URL; target: URL }[];
```
blacha, line 9: can worker threads actually pass URLS? I thought everything was meant to be a POJO for passing around?

src/commands/create-manifest/create-manifest.ts
```diff
@@ -56,10 +65,10 @@ export const commandCreateManifest = command({
 
         // Store the list of files to move in a bucket rather than the ARGO parameters
         if (actionLocation) {
-          const targetLocation = fsa.join(actionLocation, `actions/manifest-${targetHash}.json`);
+          const targetLocation = new URL(`actions/manifest-${targetHash}.json`, actionLocation.href);
```
blacha, line 68: assuming actionLocation is a URL you dont need a `.href` here ?

src/commands/stac-validate/stac.validate.ts
```diff
@@ -162,12 +162,12 @@ export const commandStacValidate = command({
           if (!checksum.startsWith('1220')) continue;
 
           let source = asset.href;
-          if (source.startsWith('./')) source = fsa.join(dirname(path), source.replace('./', ''));
+          if (source.startsWith('./')) source = new URL(dirname(path), source.replace('./', ''));
```
blacha, line 165: `new URL` would work with relative paths

src/utils/action.storage.ts
```diff
@@ -1,15 +1,17 @@
...
+
 /**
  * Store actions as JSON documents on a file system rather than passing huge JSON documents around
  *
  * Uses $ACTION_PATH if set to store the actions to a user defined location
  * This parses $ARGO_TEMPLATE looking for a `archiveLocation`
  */
-export function getActionLocation(): string | null {
-  if (process.env['ACTION_PATH']) return process.env['ACTION_PATH'];
+export async function getActionLocation(): Promise<URL | null> {
```
blacha, line 9: this doesnt need to be made async

you could also use `fsa.toUrl` or make a sync version of the URL parser

src/commands/basemaps-mapsheet/create-mapsheet.ts
l0b0: From the [Wikipedia article](https://en.wikipedia.org/wiki/File_URI_scheme#Number_of_slash_characters):

> -   The character sequence of two slash characters (//) after the string _file:_ denotes that either a hostname or the literal term _localhost_ follows,[\[3\]](https://en.wikipedia.org/wiki/File_URI_scheme#cite_note-3) although this part may be omitted entirely, or may contain an empty hostname.[\[4\]](https://en.wikipedia.org/wiki/File_URI_scheme#cite_note-4)
> -   The single slash between _**host**_ and _**path**_ denotes the start of the local-path part of the URI and must be present.[\[5\]](https://en.wikipedia.org/wiki/File_URI_scheme#cite_note-5)
> -   A valid file URI must therefore begin with either `file:/path` (no hostname), `file:///path` (empty hostname), or `file://hostname/path`.
> -   `file://path` (i.e. two slashes, without a hostname) is never correct, but is often used.
> -   Further slashes in _**path**_ separate directory names in a hierarchical system of directories and subdirectories. In this usage, the slash is a general, system-independent way of separating the parts, and in a particular host system it might be used as such in any pathname (as in Unix systems).

I thought `file://path` meant a relative path, while the final slash in file:///path indicated an absolute path, but it looks like that's wrong. We should *not* use `file://path` at all, and all paths should be absolute. I'll fix it.

src/commands/copy/copy-rpc.ts
l0b0: I assume you mean either JSON or JS object, not "[Plain old Java object](https://en.wikipedia.org/wiki/Plain_old_Java_object)". I have no idea how `WorkerRpc` works; I guess since you ask we don't have tests for this message passing?

src/commands/stac-validate/stac.validate.ts
l0b0: Not sure what you're saying. Do you mean we should just `if (source.startsWith('./')) source = new URL(dirname(path), source)`? Both `new URL('foo', './bar')` and `new URL('foo', 'bar')` complain that the second argument is not a valid URL, so I don't understand how this code ever worked, unless we were passing crazy URLs like "./http://…".

src/commands/basemaps-mapsheet/create-mapsheet.ts
```diff
@@ -72,30 +73,30 @@ export const basemapsCreateMapSheet = command({
...
-    const buf = await fsa.read(path);
+    logger.info({ path: url }, 'MapSheet:LoadFgb');
+    const buf = await fsa.read(url);
     logger.info({ config }, 'MapSheet:LoadConfig');
     const configJson = await readConfig(config);
     const mem = ConfigProviderMemory.fromJson(configJson);
 
     const rest = fgb.deserialize(buf) as FeatureCollection;
-    const featuresWritePromise = fsa.write('features.json', JSON.stringify(rest));
+    const featuresWritePromise = fsa.write(new URL(`file://${process.cwd()}/features.json`), JSON.stringify(rest));
```
blacha, line 90:
```suggestion
    const featuresWritePromise = fsa.write(new URL(fsa.toUrl('./features.json'), JSON.stringify(rest));
```

src/commands/copy/copy-rpc.ts
blacha: `worker_threads` are a mechanisim of using small scripts to add more threads to do processing, they are limited to posting messages to/from the thread using a rpc type mechanisim, `WorkerRpc` wraps the posting so it is somewhat typesafe.

I am not sure we can pass a URL between threads as its not really a easily searliazable thing, just like you cant pass a date but you can pass a ISO string.

src/commands/stac-validate/stac.validate.ts
blacha: it would be much safer to do `new URL(path, source)` which will handle all the checking for '../../foo.json` or `/foo.json`

blacha (changes requested):
These two issues would have prevented the commands from running, it is slightly concerning that they are this deep in the pull request chain and I have likely missed other spots where things are broken.

Have you run any of the commands before making this PR?

src/commands/basemaps-mapsheet/create-mapsheet.ts
```diff
@@ -72,30 +73,30 @@ export const basemapsCreateMapSheet = command({
...
-    const buf = await fsa.read(path);
+    logger.info({ path: url }, 'MapSheet:LoadFgb');
+    const buf = await fsa.read(url);
     logger.info({ config }, 'MapSheet:LoadConfig');
     const configJson = await readConfig(config);
     const mem = ConfigProviderMemory.fromJson(configJson);
 
     const rest = fgb.deserialize(buf) as FeatureCollection;
-    const featuresWritePromise = fsa.write('features.json', JSON.stringify(rest));
+    const featuresWritePromise = fsa.write(new URL(fsa.toUrl('./features.json').href), JSON.stringify(rest));
```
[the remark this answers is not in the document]
l0b0: `fsa.toUrl` doesn't exist.

src/commands/stac-validate/stac.validate.ts
```diff
@@ -109,7 +108,7 @@ export const commandStacValidate = command({
       const stacSchemas: string[] = [];
       let stacJson;
       try {
-        stacJson = await fsa.readJson<st.StacItem | st.StacCollection | st.StacCatalog>(path);
+        stacJson = await fsa.readJson<st.StacItem | st.StacCollection | st.StacCatalog>(new URL(`file://${path}`));
```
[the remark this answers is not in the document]
l0b0: Please include some hint as to why this is needed. I don't understand why this should be a file URL when the method signature is `async readJson<T>(filePath: string): Promise<T>`.

l0b0:
> These issues could have prevented the commands from running, it is slightly concerning that they are this deep in the pull request chain and I have likely missed other spots where things are broken.
> 
> Have you run any of the commands before making this PR?

No, because making heaps of manual smoke tests for each refactor is tedious, error-prone, and slow. If it's worth testing we should have automated tests for it.

src/commands/basemaps-mapsheet/create-mapsheet.ts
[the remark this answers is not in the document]
l0b0: I was using the wrong interpreter; please ignore the deleted comment.

PR #855, opened 2024-02-09
feat: Handle 1:50k imagery TDE-1014

l0b0:
#### Motivation

In addition to accepting the 50k grid size, this also produces tile names which consist of only the sheet code for 50k imagery. To do this efficiently the code irrelevant to this simpler code path have been rearranged to avoid unnecessary computation.

#### Checklist

- [x] Tests updated
- [ ] Docs updated (undocumented)
- [x] Issue linked in Title

### DIFF
--- a/src/commands/tileindex-validate/__test__/tileindex.validate.test.ts
+++ b/src/commands/tileindex-validate/__test__/tileindex.validate.test.ts
@@ -44,6 +44,11 @@ describe('getTileName', () => {
     assert.equal(convertTileName('CH11_1000_1010', 10000), 'CH11_10000_0101');
     assert.equal(convertTileName('CH11_1000_1001', 10000), 'CH11_10000_0101');
   });
+  it('should get correct parent tile 1:50k', () => {
+    assert.equal(convertTileName('AT24_50000_0101', 50000), 'AT24');
+    assert.equal(convertTileName('AT25_50000_0101', 50000), 'AT25');
+    assert.equal(convertTileName('CK08_50000_0101', 50000), 'CK08');
+  });
 });
 describe('tiffLocation', () => {
   it('get location from tiff', async () => {
@@ -230,7 +235,7 @@ describe('GridSizeFromString', () => {
   it('should throw error when converting invalid grid size', async () => {
     await assert.rejects(
       GridSizeFromString.from('-1'),
-      new Error('Invalid grid size "-1"; valid values: "10000", "5000", "2000", "1000", "500"'),
+      new Error('Invalid grid size "-1"; valid values: "50000", "10000", "5000", "2000", "1000", "500"'),
     );
   });
 });
--- a/src/utils/mapsheet.ts
+++ b/src/utils/mapsheet.ts
@@ -119,6 +119,19 @@ export const MapSheet = {
       height: 0,
       bbox: [0, 0, 0, 0],
     };
+
+    const mapSheetOffset = MapSheet.offset(sheetCode);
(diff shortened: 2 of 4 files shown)


src/commands/tileindex-validate/__test__/tileindex.validate.test.ts
```diff
@@ -230,7 +235,7 @@ describe('GridSizeFromString', () => {
   it('should throw error when converting invalid grid size', async () => {
     await assert.rejects(
       GridSizeFromString.from('-1'),
-      new Error('Invalid grid size "-1"; valid values: "10000", "5000", "2000", "1000", "500"'),
+      new Error('Invalid grid size "-1"; valid values: "50000", "10000", "5000", "2000", "1000", "500"'),
```
paulfouquet, line 238: Is it possible to list the values in [gridSizes](https://github.com/linz/argo-tasks/blob/9a7ba359dee9fbd23586ea8a45d72fd6b0c06af7/src/utils/mapsheet.ts#L59)?

l0b0: This is a tricky one, in tests. Basically, how much of the test code should be dynamic? If we generate this expected string, the test doesn't have to change when the list changes. But there are issues with replacing this line with

```typescript
new Error(`Invalid grid size "-1"; valid values: "${gridSizes.join('", "')}"`)
```

or even

```typescript
new Error(`Invalid grid size "${value}"; valid values: "${gridSizes.join('", "')}"`)
```

(the exact [production code](https://github.com/linz/argo-tasks/blob/18576e6a246c46dd65f26dc2a24c5531af9703d6/src/commands/tileindex-validate/tileindex.validate.ts#L111)). It's much harder to tell whether the expected error message is *actually* what we expect. For example, what if we passed the integer literal `1e5` or `100_000` into `GridSizeFromString.from`? The error message would say "100000", not the same as the original literal. Is this what we want? Depends on the application. Maybe we'd want "100,000" in the error message, but only for English, where the thousands separator is usually comma. We're not testing that here, because it's just not a very important use case compared to all the other stuff, but for any code we copy from production we're basically saying we don't care about testing that piece of code. And what if there's a bug in all that mess of nested code and strings? By using a literal as the expected value we're checking that the literal is actually what we want, rather than testing that the same code, executed within the test context and production context produce the same result.

paulfouquet: Seems too complex. Allgood then!

paulfouquet (dismissed):

src/utils/mapsheet.ts
```diff
@@ -117,11 +121,11 @@ export const MapSheet = {
...
     if (out.gridSize === 500) {
-      out.y = Number(match[3]?.slice(0, 3));
-      out.x = Number(match[3]?.slice(3));
+      out.y = Number(match?.groups?.['tileId']?.slice(0, 3));
+      out.x = Number(match?.groups?.['tileId']?.slice(3));
     } else {
-      out.y = Number(match[3]?.slice(0, 2));
-      out.x = Number(match[3]?.slice(2));
+      out.y = Number(match?.groups?.['tileId']?.slice(0, 2));
+      out.x = Number(match?.groups?.['tileId']?.slice(2));
```
amfage, line 128: I just realised we need a case for the 50k imagery as this returns, as `out.x` and `out.y` will be NaN for the 50k imagery.

paulfouquet (approved):

amfage (approved):

PR #860, opened 2024-02-11
fix: Lints TDE-1030

l0b0:
#### Motivation

Simplify dealing with the code.

#### Checklist

- [ ] Tests updated (N/A)
- [ ] Docs updated (N/A)
- [x] Issue linked in Title


paulfouquet (approved):

src/commands/path/__test__/generate.path.test.ts
```diff
@@ -161,13 +161,13 @@ describe('gsd', () => {
...
-      extractGsd(TiffNoGsd), Error;
-    });
+      extractGsd(TiffNoGsd);
+    }, Error);
   });
 });
 
 describe('category', async () => {
-  it('Should return category', async () => {
+  await it('Should return category', async () => {
```
blacha, line 170: why are these tests awaited?

src/commands/path/__test__/generate.path.test.ts
```diff
@@ -111,8 +111,8 @@ describe('GeneratePathHistoricImagery', () => {
       epsg: 2193,
     };
     assert.throws(() => {
```
blacha, line 113: Having just `Error` as the throws checker isn't really adding anything, could it be improved with matching the error message?

otherwise you could drop the Error `assert.throws(() => generatePath(metadata))` would be a similar enough check.

l0b0: > Having just `Error` as the throws checker isn't really adding anything, could it be improved with matching the error message?

Absolutely, but I don't feel like that's relevant for this PR.

> otherwise you could drop the Error `assert.throws(() => generatePath(metadata))` would be a similar enough check.

:shrug: I guess that weakens the test a little bit, since it won't check the error type anymore. Is this worth doing?

src/commands/path/__test__/generate.path.test.ts
blacha: your IDE is incorrectly detecting these, these should not be awaited.

blacha (approved):

PR #874, opened 2024-02-18
build: add a format script based on prettier

paulfouquet:
#### Motivation

Before pushing some changes, I want my modifications to have a correct format based on the LINZ prettier config.

#### Modification

Add a `format` script that can be run manually using `npm run format`

#### Checklist

- [ ] Tests updated N/A
- [ ] Docs updated N/A
- [ ] Issue linked in Title N/A


package.json
```diff
@@ -22,6 +22,7 @@
   "scripts": {
     "build": "tsc",
     "lint": "npx eslint . --ignore-path .gitignore",
+    "format": "npx prettier . -w",
```
l0b0, line 25: Options should come before paths, in basically all \*nix commands. Also, it's probably useful to use long options for readability:

```suggestion
    "format": "npx prettier --write .",
```

paulfouquet:
There are a couple of issues with prettier :thinking: 
- not adding the newline at the end of the files
- in github actions workflows, `${{env.MY_VAR}}` should be formatted to `${{ env.MY_VAR }}`, but it's not

PR #877, opened 2024-02-19
fix: tileindex-validate errors should not been thrown asap TDE-1013

paulfouquet:
#### Motivation

In order to have a better overview of the errors, errors should be collected and logged at the end rathen than thrown ASAP.

#### Modification

- Avoid throwing error as soon as they're detected
- Couple of refactors to simplify the code

#### Checklist

- [ ] Tests updated N/A
- [ ] Docs updated N/A
- [x] Issue linked in Title

### DIFF
--- a/src/commands/tileindex-validate/tileindex.validate.ts
+++ b/src/commands/tileindex-validate/tileindex.validate.ts
@@ -223,10 +223,10 @@ export const commandTileIndexValidate = command({
     if (args.validate) {
       let validationFailed = false;
       for (const tiff of locations) {
-        const ret = validateTiffAlignment(tiff);
-        if (ret === true) continue;
-        logger.error({ reason: ret.message, source: tiff.source }, 'TileInvalid:Validation:Failed');
-        validationFailed = true;
+        const valid = validateTiffAlignment(tiff);
+        if (valid === false) {
+          validationFailed = true;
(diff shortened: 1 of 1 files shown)


src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -223,10 +223,10 @@ export const commandTileIndexValidate = command({
...
       let validationFailed = false;
       for (const tiff of locations) {
-        const ret = validateTiffAlignment(tiff);
-        if (ret === true) continue;
-        logger.error({ reason: ret.message, source: tiff.source }, 'TileInvalid:Validation:Failed');
-        validationFailed = true;
+        const valid = validateTiffAlignment(tiff);
+        if (valid === false) {
+          validationFailed = true;
+        }
```
l0b0, line 229: I'd recommend using a single variable since we don't need `valid` except to update `validationFailed`. I'd also recommend using positive words in general, to avoid any double negatives. In other words, `validationFailed` could be something like `allValid`, and you could do something like `allValid = allValid && validateTiffAlignment(tiff);`

l0b0 (dismissed):

l0b0 (approved):

MDavidson17 (approved):

PR #890, opened 2024-02-26
feat: detects and error if the tiff is not 8 bits TDE-895

paulfouquet:
#### Motivation

The standardising workflow is only processing 8 bits TIFFs at the moment.

#### Modification

In order to exclude any other TIFF than 8 bits ones, tileindex-validate throw an error if one TIFF is not 8 bits.
~~TODO: unit test - need to mock the TIFF tag BitsPerSample~~ Added some test files.

#### Checklist

- [ ] Tests updated
- [x] Docs updated
- [x] Issue linked in Title


l0b0 (commented):
Would you be OK to split the package upgrade to a separate PR? It clutters up the diff.

src/commands/tileindex-validate/__test__/tileindex.validate.test.ts
```diff
@@ -276,3 +278,14 @@ describe('GridSizeFromString', () => {
     );
   });
 });
+
+describe('is8BitsTiff', () => {
+  it('should be a 8bits Tiff', async () => {
```
l0b0, line 283: - "TIFF" is an abbreviation, so it should be all upper case
- "8-bit" or "8 bit" are the normal ways to format that part of the string.

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -416,3 +418,25 @@ export function getTileName(x: number, y: number, gridSize: GridSize): string {
...
+
+/**
+ * Verify if a TIFF contains only 8 bits bands.
+ *
+ * @param tiff
+ * @returns true if 8 bits TIFF
+ */
+export async function is8BitsTiff(tiff: Tiff): Promise<boolean> {
+  const baseImage = tiff.images[0];
+  if (baseImage === undefined) throw new Error("Can't get base image");
```
l0b0, line 430: Do we need more info to debug this? For example, when this is triggered, would the user trivially understand which TIFF this applied to?

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -46,6 +46,8 @@ export const TiffLoader = {
     for (const prom of promises) {
       // All the errors are logged above so just throw the first error
       if (prom.status === 'rejected') throw new Error('Tiff loading failed: ' + String(prom.reason));
+      // We are processing only 8 bits Tiff for now
+      if ((await is8BitsTiff(prom.value)) === false) throw new Error('Tiff is not a 8 bits TIFF.');
```
l0b0, line 50: It looks like we don't need the return value other than to throw this error. In that case we could throw the error inside `is8BitsTiff` and return nothing.

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -416,3 +418,25 @@ export function getTileName(x: number, y: number, gridSize: GridSize): string {
...
   return `${sheetCode}_${gridSize}_${tileId}`;
 }
+
+/**
+ * Verify if a TIFF contains only 8 bits bands.
+ *
+ * @param tiff
+ * @returns true if 8 bits TIFF
+ */
+export async function is8BitsTiff(tiff: Tiff): Promise<boolean> {
```
l0b0, line 428: Would it make sense to change this so that we won't have to change the name once we support other bit depths? That is, would a name like `validateBitDepth` be future-proof?

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -416,3 +418,25 @@ export function getTileName(x: number, y: number, gridSize: GridSize): string {
...
+  const baseImage = tiff.images[0];
+  if (baseImage === undefined) throw new Error("Can't get base image");
+
+  const bitsPerSample = await baseImage.fetch(TiffTag.BitsPerSample);
+  if (bitsPerSample == null) {
+    throw new Error('Failed to extract band information from : ' + tiff.source.url);
+  }
+
+  for (let i = 0; i < bitsPerSample.length; i++) {
+    if (bitsPerSample[i] !== 8) return false;
```
l0b0, line 438: We could use [`every`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) here.

src/commands/tileindex-validate/tileindex.validate.ts
paulfouquet: That is something I generally do when the need of supporting something else comes.

paulfouquet:
> Would you be OK to split the package upgrade to a separate PR? It clutters up the diff.

Do you mean merging the package upgrade with a separate PR before having this PR? 
That is an option but having the reason behind the upgrade in the same PR is somewhat useful, and I don't feel the refactor is that big?

l0b0 (dismissed):

l0b0 (approved):

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -416,3 +418,21 @@ export function getTileName(x: number, y: number, gridSize: GridSize): string {
...
+  const baseImage = tiff.images[0];
+  if (baseImage === undefined) throw new Error(`Can't get base image for ${tiff.source.url}`);
+
+  const bitsPerSample = await baseImage.fetch(TiffTag.BitsPerSample);
+  if (bitsPerSample == null) {
+    throw new Error(`Failed to extract band information from ${tiff.source.url}`);
+  }
+
+  if (!bitsPerSample.every((currentNumberBits) => currentNumberBits === 8))
+    throw new Error(`${tiff.source.url} is not a 8 bits TIFF`);
```
blacha, line 437: if things are multi lined its generally a good idea to include brackets.

```suggestion
  if (!bitsPerSample.every((currentNumberBits) => currentNumberBits === 8)) {
    throw new Error(`${tiff.source.url} is not a 8 bits TIFF`);
  }
```

l0b0: This seems like the sort of things our linter should enforce.

blacha: It should be covered by lint I'm not sure why its not enforcing brackets here.

l0b0 (approved):

blacha (approved):

PR #915, opened 2024-03-12
fix(basemaps): Fix the event category is not parsed and populated in the pull request. BM-982

Wentao-Kuang:
#### Motivation

The `Event` category are populated as `New Aerial Photos` when creating pull request.

#### Modification

- the `parseCategory` not parsing the `event` and `scanned` category. Just added them to fix.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/src/commands/basemaps-github/create-pr.ts
+++ b/src/commands/basemaps-github/create-pr.ts
@@ -8,7 +8,7 @@ import { StacCollection } from 'stac-ts';
 import { CliInfo } from '../../cli.info.js';
 import { logger } from '../../log.js';
 import { registerCli, verbose } from '../common.js';
-import { Category, MakeCogGithub, parseCategory } from './make.cog.github.js';
+import { Category, MakeCogGithub } from './make.cog.github.js';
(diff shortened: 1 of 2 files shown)


l0b0 (approved):

l0b0:
Nit: Would it be possible to use an object lookup, as in `Category[c] || Category.Other`?

Wentao-Kuang:
> Nit: Would it be possible to use an object lookup, as in `Category[c] || Category.Other`?

I think we probably can just delete this `parseCategory` function which is using for the old command line tool. The new `cmd-ts` already parsed it as one of the `Category`. So we can using the `arg.category` directly as they have to be valid ones.
 https://github.com/linz/argo-tasks/blob/59e0ab4b8db4c1864734ca31450f727152afd1aa/src/commands/basemaps-github/create-pr.ts#L72

l0b0 (approved):

blacha (approved):

PR #933, opened 2024-03-21
feat(basemaps): Update the create-pr cli to add new individual vector tileset. BM-992

Wentao-Kuang:
#### Motivation

We need to run another cron workflow to generate individual vector map output for the nz road addressing layer. So we need to allow the create-pr command to be able to add create pull request for it.

#### Modification

- Remove the hard coded topographic vector tileset logic
- Update the vector to parse the info from vector target and prepare new tileset for it.
- Update `getContent` to return null for Not Found

#### Checklist

- [ ] Tests updated - No related tests.
- [ ] Docs updated - Doc no need to update
- [x] Issue linked in Title

### DIFF
--- a/src/commands/basemaps-github/create-pr.ts
+++ b/src/commands/basemaps-github/create-pr.ts
@@ -13,6 +13,14 @@ import { Category, MakeCogGithub } from './make.cog.github.js';
 const validTargetBuckets: Set<string> = new Set(['linz-basemaps', 'linz-basemaps-staging']);
 const validSourceBuckets: Set<string> = new Set(['nz-imagery', 'linz-imagery']);
 
+async function validateBucket(bucket: string, validBuckets: Set<string>): Promise<void> {
+  // Validate the target information
+  logger.info({ bucket }, 'CreatePR: Valid the target s3 bucket');
+  if (bucket == null || !validBuckets.has(bucket)) {
+    throw new Error(`Invalid s3 bucket ${bucket} from the target.`);
+  }
+}
+
 async function parseRasterTargetInfo(
   target: string,
   individual: boolean,
@@ -24,11 +32,7 @@ async function parseRasterTargetInfo(
   const epsg = Epsg.tryGet(Number(splits[1]));
   const name = splits[2];
 
-  //Validate the target information
-  logger.info({ bucket }, 'CreatePR: Valid the target s3 bucket');
-  if (bucket == null || !validTargetBuckets.has(bucket)) {
-    throw new Error(`Invalid s3 bucket ${bucket} from the target ${target}.`);
-  }
+  await validateBucket(bucket, validTargetBuckets);
 
   if (epsg == null || name == null) throw new Error(`Invalid target ${target} to parse the epsg and imagery name.`);
   const collectionPath = fsa.join(target, 'collection.json');
@@ -42,10 +46,8 @@ async function parseRasterTargetInfo(
   if (source == null) throw new Error(`Failed to get source url from collection.json.`);
   const sourceUrl = new URL(source);
   const sourceBucket = sourceUrl.hostname;
-  logger.info({ bucket: sourceBucket }, 'CreatePR: Validate the source s3 bucket');
-  if (sourceBucket == null || !validSourceBuckets.has(sourceBucket)) {
-    throw new Error(`Invalid s3 bucket ${sourceBucket} from the source ${sourceUrl}.`);
-  }
+  await validateBucket(sourceBucket, validSourceBuckets);
+
   // Try to get the region for individual layers
   let region;
   if (individual) {
@@ -74,11 +76,7 @@ async function parseVectorTargetInfo(target: string): Promise<{ name: string; ti
   const epsg = Epsg.tryGet(Number(splits[2]));
   const name = splits[3];
 
-  // Validate the target information
-  logger.info({ bucket }, 'CreatePR: Valid the target s3 bucket');
-  if (bucket == null || !validTargetBuckets.has(bucket)) {
-    throw new Error(`Invalid s3 bucket ${bucket} from the target ${target}.`);
-  }
+  await validateBucket(bucket, validTargetBuckets);
 
   if (epsg == null || name == null) throw new Error(`Invalid target ${target} to parse the epsg and imagery name.`);
   if (epsg !== Epsg.Google) throw new Error(`Unsupported epsg code ${epsg.code} for vector map.`);
--- a/src/commands/basemaps-github/create-pr.ts
+++ b/src/commands/basemaps-github/create-pr.ts
@@ -61,7 +61,6 @@ async function parseRasterTargetInfo(
   return { name: standardizeLayerName(name), epsg: epsg.code, title, region };
 }
 
-
 /**
  * Pass vector target location with the following format
  * s3://linz-basemaps-staging/vector/3857/53382-nz-roads-addressing/01HSF04SG9M1P3V667A4NZ1MN8/53382-nz-roads-addressing.tar.co
--- a/src/commands/basemaps-github/create-pr.ts
+++ b/src/commands/basemaps-github/create-pr.ts
@@ -16,7 +16,7 @@ const validSourceBuckets: Set<string> = new Set(['nz-imagery', 'linz-imagery']);
 function assertValidBucket(bucket: string, validBuckets: Set<string>): void {
   // Validate the target information
   logger.info({ bucket }, 'CreatePR: Valid the target s3 bucket');
-  if (bucket == null || !validBuckets.has(bucket)) {
+  if (!validBuckets.has(bucket)) {
     throw new Error(`Invalid s3 bucket ${bucket} from the target.`);
   }
 }
--- a/src/commands/basemaps-github/create-pr.ts
+++ b/src/commands/basemaps-github/create-pr.ts
@@ -13,7 +13,7 @@ import { Category, MakeCogGithub } from './make.cog.github.js';
 const validTargetBuckets: Set<string> = new Set(['linz-basemaps', 'linz-basemaps-staging']);
 const validSourceBuckets: Set<string> = new Set(['nz-imagery', 'linz-imagery']);
 
-async function parseTargetInfo(
+async function parseRasterTargetInfo(
   target: string,
   individual: boolean,
 ): Promise<{ name: string; title: string; epsg: EpsgCode; region: string | undefined }> {
@@ -59,6 +59,33 @@ async function parseTargetInfo(
   }
 
   return { name: standardizeLayerName(name), epsg: epsg.code, title, region };
+}
+
+
+/**
+ * Pass vector target location with the following format
+ * s3://linz-basemaps-staging/vector/3857/53382-nz-roads-addressing/01HSF04SG9M1P3V667A4NZ1MN8/53382-nz-roads-addressing.tar.co
+ * s3://linz-basemaps-staging/vector/3857/topographic/01HSF04SG9M1P3V667A4NZ1MN8/topographic.tar.co
+ */
+async function parseVectorTargetInfo(target: string): Promise<{ name: string; title: string; epsg: EpsgCode }> {
+  logger.info({ target }, 'CreatePR: Get the layer information from target');
+  const url = new URL(target);
+  const bucket = url.hostname;
+  const splits = url.pathname.split('/');
+  const epsg = Epsg.tryGet(Number(splits[2]));
+  const name = splits[3];
+
+  //Validate the target information
+  logger.info({ bucket }, 'CreatePR: Valid the target s3 bucket');
+  if (bucket == null || !validTargetBuckets.has(bucket)) {
+    throw new Error(`Invalid s3 bucket ${bucket} from the target ${target}.`);
+  }
+
+  if (epsg == null || name == null) throw new Error(`Invalid target ${target} to parse the epsg and imagery name.`);
+  if (epsg !== Epsg.Google) throw new Error(`Unsupported epsg code ${epsg.code} for vector map.`);
+  // Try to get the region for individual layers
+
+  return { name: name, epsg: epsg.code, title: name };
 }
 
 export const CommandCreatePRArgs = {
@@ -113,12 +140,15 @@ export const basemapsCreatePullRequest = command({
     const layer: ConfigLayer = { name: '', title: '' };
     let region;
     if (args.vector) {
-      layer.name = 'topographic';
-      layer.title = 'Topographic';
-      layer[3857] = targets[0];
+      for (const target of targets) {
+        const info = await parseVectorTargetInfo(target);
+        layer.name = info.name;
+        layer.title = info.title;
+        layer[info.epsg] = target;
+      }
     } else {
       for (const target of targets) {
-        const info = await parseTargetInfo(target, args.individual);
+        const info = await parseRasterTargetInfo(target, args.individual);
         layer.name = info.name;
         layer.title = info.title;
         layer[info.epsg] = target;
@@ -130,7 +160,7 @@ export const basemapsCreatePullRequest = command({
     if (layer.name === '' || layer.title === '') throw new Error('Failed to find the imagery name or title.');
 
     const git = new MakeCogGithub(layer.name, args.repository);
-    if (args.vector) await git.updateVectorTileSet('topographic', layer);
+    if (args.vector) await git.updateVectorTileSet(layer.name, layer);
     else await git.updateRasterTileSet('aerial', layer, category, region);
   },
 });
--- a/src/commands/basemaps-github/create-pr.ts
+++ b/src/commands/basemaps-github/create-pr.ts
@@ -13,7 +13,7 @@ import { Category, MakeCogGithub } from './make.cog.github.js';
 const validTargetBuckets: Set<string> = new Set(['linz-basemaps', 'linz-basemaps-staging']);
 const validSourceBuckets: Set<string> = new Set(['nz-imagery', 'linz-imagery']);
 
-async function validateBucket(bucket: string, validBuckets: Set<string>): Promise<void> {
+function assertValidBucket(bucket: string, validBuckets: Set<string>): void {
   // Validate the target information
   logger.info({ bucket }, 'CreatePR: Valid the target s3 bucket');
   if (bucket == null || !validBuckets.has(bucket)) {
@@ -32,7 +32,7 @@ async function parseRasterTargetInfo(
   const epsg = Epsg.tryGet(Number(splits[1]));
   const name = splits[2];
 
-  await validateBucket(bucket, validTargetBuckets);
+  assertValidBucket(bucket, validTargetBuckets);
 
   if (epsg == null || name == null) throw new Error(`Invalid target ${target} to parse the epsg and imagery name.`);
   const collectionPath = fsa.join(target, 'collection.json');
@@ -46,7 +46,7 @@ async function parseRasterTargetInfo(
   if (source == null) throw new Error(`Failed to get source url from collection.json.`);
   const sourceUrl = new URL(source);
   const sourceBucket = sourceUrl.hostname;
-  await validateBucket(sourceBucket, validSourceBuckets);
+  assertValidBucket(sourceBucket, validSourceBuckets);
 
   // Try to get the region for individual layers
   let region;
@@ -76,7 +76,7 @@ async function parseVectorTargetInfo(target: string): Promise<{ name: string; ti
   const epsg = Epsg.tryGet(Number(splits[2]));
   const name = splits[3];
 
-  await validateBucket(bucket, validTargetBuckets);
+  assertValidBucket(bucket, validTargetBuckets);
 
   if (epsg == null || name == null) throw new Error(`Invalid target ${target} to parse the epsg and imagery name.`);
   if (epsg !== Epsg.Google) throw new Error(`Unsupported epsg code ${epsg.code} for vector map.`);
--- a/src/commands/basemaps-github/create-pr.ts
+++ b/src/commands/basemaps-github/create-pr.ts
@@ -74,7 +74,7 @@ async function parseVectorTargetInfo(target: string): Promise<{ name: string; ti
   const epsg = Epsg.tryGet(Number(splits[2]));
   const name = splits[3];
 
-  //Validate the target information
+  // Validate the target information
   logger.info({ bucket }, 'CreatePR: Valid the target s3 bucket');
   if (bucket == null || !validTargetBuckets.has(bucket)) {
     throw new Error(`Invalid s3 bucket ${bucket} from the target ${target}.`);
--- a/src/commands/basemaps-github/make.cog.github.ts
+++ b/src/commands/basemaps-github/make.cog.github.ts
@@ -37,6 +37,13 @@ export const DefaultCategorySetting: Record<Category, CategorySetting> = {
   [Category.Event]: {},
 };
 
+interface StacLinkLds extends StacLink {
+  'lds:id': string;
+  'lds:name': string;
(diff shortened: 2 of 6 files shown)


src/commands/basemaps-github/make.cog.github.ts
```diff
@@ -170,10 +171,24 @@ export class MakeCogGithub {
     logger.info({ imagery: this.imagery }, 'GitHub: Get the master TileSet config file');
     const tileSetPath = fsa.joinAll('config', 'tileset', `${filename}.json`);
     const tileSetContent = await gh.getContent(tileSetPath);
-    const tileSet = JSON.parse(tileSetContent) as ConfigTileSetVector;
-    const newTileSet = await this.prepareVectorTileSetConfig(layer, tileSet);
+    let newTileSet;
```
blacha, line 174: this is typed as a any, generally we should not be using `let` if we can avoid it, could you refactor it so the let is not needed?

src/commands/basemaps-github/create-pr.ts
```diff
@@ -61,6 +61,32 @@ async function parseTargetInfo(
...
 }
 
+/**
+ * Pass vector target location with the following format
+ * s3://linz-basemaps-staging/vector/3857/53382-nz-roads-addressing/01HSF04SG9M1P3V667A4NZ1MN8/53382-nz-roads-addressing.tar.co
+ * s3://linz-basemaps-staging/vector/3857/topographic/01HSF04SG9M1P3V667A4NZ1MN8/topographic.tar.co
+ */
+async function parseVectorTargetInfo(target: string): Promise<{ name: string; title: string; epsg: EpsgCode }> {
+  logger.info({ target }, 'CreatePR: Get the layer information from target');
+  const url = new URL(target);
```
blacha, line 71: Could we read all this information from the collection.json/item.json rather than trying to parse values from URLs?

src/commands/basemaps-github/create-pr.ts
```diff
@@ -61,6 +61,32 @@ async function parseTargetInfo(
...
+  const bucket = url.hostname;
+  const splits = url.pathname.split('/');
+  const epsg = Epsg.tryGet(Number(splits[2]));
+  const name = splits[3];
```
l0b0, line 75: A lot of this functionality looks like it's a copy of `parseRasterTargetInfo`. Would it be possible to extract a method to reuse most of it?

src/utils/github.ts
```diff
@@ -94,14 +94,15 @@ export class GithubApi {
...
+  async getContent(path: string): Promise<string | undefined> {
     logger.info({ path }, 'GitHub API: Get Content');
     const response = await this.octokit.rest.repos.getContent({ owner: this.owner, repo: this.repo, path });
     if (!this.isOk(response.status)) throw new Error('Failed to get aerial TileSet config.');
     if ('content' in response.data) {
       return Buffer.from(response.data.content, 'base64').toString();
     } else {
-      throw new Error(`Unable to find the content from path ${path}.`);
+      logger.info({ path }, 'Unable to find the content from path');
+      return;
```
l0b0, line 105: This change doesn't seem to be explained in the PR description or commit messages.

src/commands/basemaps-github/create-pr.ts
Wentao-Kuang: I had a look of ETL stac files, the only thing I think is useful here is the item.properties['proj:epsg'], So I thought just parse url should be enough for now.

https://github.com/linz/basemaps-team/blob/a791140c3d76ef0a7c4967082de759cb89f51876/packages/ETL/src/stac/stac.ts#L15

src/commands/basemaps-github/create-pr.ts
```diff
@@ -13,7 +13,15 @@ import { Category, MakeCogGithub } from './make.cog.github.js';
 const validTargetBuckets: Set<string> = new Set(['linz-basemaps', 'linz-basemaps-staging']);
 const validSourceBuckets: Set<string> = new Set(['nz-imagery', 'linz-imagery']);
 
-async function parseTargetInfo(
+async function validateBucket(bucket: string, validBuckets: Set<string>): Promise<void> {
```
blacha, line 16: nit: this is a assertion function as it throws on error

`assertValidBucket(..)`

blacha: it also should not be async

Wentao-Kuang: Thanks, just updated.

src/commands/basemaps-github/make.cog.github.ts
```diff
@@ -161,6 +162,83 @@ export class MakeCogGithub {
...
+  }
+
+  /**
+   * Prepare and create pull request for the aerial tileset config
+   */
+  async diffVectorUpdate(layer: ConfigLayer, existingTileSet?: ConfigTileSetVector): Promise<string> {
+    let md = '';
+    // Vector layer only support for 3857
+    if (layer[3857] == null) return md;
+    const newCollectionPath = this.getCollectionUrl(layer[3857]);
```
blacha, line 184: `new URL("collection.json", layer[3857])`

src/commands/basemaps-github/make.cog.github.ts
```diff
@@ -161,6 +162,83 @@ export class MakeCogGithub {
...
+  getCollectionUrl(target: string): string {
+    const path = target.substring(0, target.lastIndexOf('/'));
+    return fsa.join(path, 'collection.json');
+  }
+
+  /**
+   * Prepare and create pull request for the aerial tileset config
+   */
+  async diffVectorUpdate(layer: ConfigLayer, existingTileSet?: ConfigTileSetVector): Promise<string> {
+    let md = '';
```
blacha, line 181: `const output: string[] = [];`

src/commands/basemaps-github/make.cog.github.ts
```diff
@@ -161,6 +162,83 @@ export class MakeCogGithub {
...
+    // Log all the new inserts for new tileset
+    if (existingTileSet == null) {
+      md += `New TileSet ts_${layer.name} with layer ${layer.name}.\n`;
+      for (const l of ldsLayers) {
+        md += `🟩 ${l['lds:name']} - version: ${l['lds:version']} features: ${l['lds:feature_count']}`;
+      }
+      return md;
+    }
+
+    // Compare the different of existing tileset
```
blacha, line 198: could we add a comment saying there is generally only one layer here so this isnt going to fetch very much.

src/commands/basemaps-github/make.cog.github.ts
```diff
@@ -159,6 +162,89 @@ export class MakeCogGithub {
...
+    for (const l of existingTileSet.layers) {
+      if (l[3857] == null) continue;
+      if (l.name !== layer.name) continue;
+      changes.push(`Update for TileSet ${existingTileSet.id} layer ${layer.name}.`);
+      const existingCollectionPath = new URL('collection.json', l[3857]).href;
+      const existingCollection = await fsa.readJson<StacCollection>(existingCollectionPath);
+      if (existingCollection == null) {
+        throw new Error(`Failed to get target collection json from ${existingCollectionPath}.`);
+      }
+      const existingLdsLayers = existingCollection.links.filter((f) => f.rel === 'lds:layer');
```
blacha, line 222: this should be a map
```suggestion
      const existingLdsLayers = new Map<string, StacLink>();
      for (const item of existingCollection.links) {
       if (item.rel === 'lds:layer') existingLdsLayers.set(item['lds:id'], item)
      };
```

src/commands/basemaps-github/make.cog.github.ts
```diff
@@ -159,6 +162,89 @@ export class MakeCogGithub {
...
+      for (const l of ldsLayers) {
+        const existingLayer = existingLdsLayers.find((f) => f['lds:id'] === l['lds:id']);
+        if (existingLayer == null) {
```
blacha, line 225:
```
const existing = exlistingLdsLayers.get(f['lds:id'])
// Switches changes to (string|null)[]
changes.push(this.getVectorChanges(l, existingLayer));


change.filter(f => f!=null).join('\n')
```

src/commands/basemaps-github/make.cog.github.ts
```diff
@@ -159,6 +162,89 @@ export class MakeCogGithub {
...
+    }
+
+    // Remove Layer
+    if (newLayer == null && existingLayer != null) {
+      return `🟥 ${existingLayer['lds:name']} features: -${existingLayer['lds:feature_count']}`;
+    }
+
+    // Update Layer
+    if (newLayer != null && existingLayer != null) {
+      const featureChange = Number(newLayer['lds:feature_count']) - Number(existingLayer['lds:feature_count']);
```
blacha, line 178: if (newLayer.version === oldLayer.version)
 if (featureChange === 0) return null;
 // Giant red box `🟥🟥🟥🟥 Feature Change detected 🟥🟥🟥🟥`

src/commands/basemaps-github/create-pr.ts
Wentao-Kuang: I extracted the duplicated logic into `assertValidBucket` function.

src/commands/basemaps-github/make.cog.github.ts
```diff
@@ -159,6 +162,97 @@ export class MakeCogGithub {
...
+    if (newLayer != null && existingLayer == null) {
+      return `🟩 ${newLayer['lds:name']} - version: ${newLayer['lds:version']} features: ${newLayer['lds:feature_count']}`;
+    }
+
+    // Remove Layer
+    if (newLayer == null && existingLayer != null) {
+      return `🟥 ${existingLayer['lds:name']} features: -${existingLayer['lds:feature_count']}`;
+    }
+
+    return;
```
blacha, line 197:
```suggestion
```

src/commands/basemaps-github/make.cog.github.ts
```diff
@@ -159,6 +162,97 @@ export class MakeCogGithub {
...
+      if (l.name !== layer.name) continue;
+      changes.push(`Update for TileSet ${existingTileSet.id} layer ${layer.name}.`);
+      const existingCollectionPath = new URL('collection.json', l[3857]).href;
+      const existingCollection = await fsa.readJson<StacCollection>(existingCollectionPath);
+      if (existingCollection == null) {
+        throw new Error(`Failed to get target collection json from ${existingCollectionPath}.`);
+      }
+
+      // Prepare existing lds layers as map
+      const existingLdsLayers = new Map<unknown, StacLink>();
```
blacha, line 233:
```suggestion
      const existingLdsLayers = new Map<string, StacLink>();
```

do you then need to type the `(f) => f.rel === 'lds:layer') as StacLinkLds;` ?

src/commands/basemaps-github/make.cog.github.ts
Wentao-Kuang: This function return `string | undefined` now, we need a return at end.

src/commands/basemaps-github/make.cog.github.ts
Wentao-Kuang: Yeah, we might need add a type for this other wise all the values are unknown.

src/commands/basemaps-github/make.cog.github.ts
```diff
@@ -159,6 +162,97 @@ export class MakeCogGithub {
...
+      for (const item of existingCollection.links) {
+        if (item.rel === 'lds:layer') existingLdsLayers.set(item['lds:id'], item);
+      }
+
+      // Find layer updates
+      for (const l of ldsLayers) {
+        const existingLayer = existingLdsLayers.get(l['lds:id']);
```
blacha, line 240: This should remove from `existingLdsLayers` so if there are duplicates it would be detected, it would also simplify the next loop as removals are just is there anything in `existingLdsLayers`

src/commands/basemaps-github/make.cog.github.ts
```diff
@@ -159,6 +169,95 @@ export class MakeCogGithub {
...
+        changes.push(this.getVectorChanges(l, existingLayer));
+        if (existingLayer != null) existingLdsLayers.delete(l['lds:id']);
+      }
+
+      // Remove the layers that not deleted from existingLdsLayers
+      for (const id of existingLdsLayers.keys()) {
+        changes.push(this.getVectorChanges(undefined, existingLdsLayers.get(id)));
+      }
```
blacha, line 255:
```suggestion
      for (const layer of existingLdsLayers.values()) {
        changes.push(this.getVectorChanges(undefined, layer));
      }
```

blacha (approved):

src/commands/basemaps-github/create-pr.ts
```diff
@@ -13,7 +13,15 @@ import { Category, MakeCogGithub } from './make.cog.github.js';
 const validTargetBuckets: Set<string> = new Set(['linz-basemaps', 'linz-basemaps-staging']);
 const validSourceBuckets: Set<string> = new Set(['nz-imagery', 'linz-imagery']);
 
-async function parseTargetInfo(
+function assertValidBucket(bucket: string, validBuckets: Set<string>): void {
+  // Validate the target information
+  logger.info({ bucket }, 'CreatePR: Valid the target s3 bucket');
+  if (bucket == null || !validBuckets.has(bucket)) {
```
l0b0, line 19: Do we need to check for nullity here?

src/utils/github.ts
```diff
@@ -94,15 +94,21 @@ export class GithubApi {
...
+        if ('content' in response.data) {
+          return Buffer.from(response.data.content, 'base64').toString();
+        } else {
+          throw new Error(`Unable to find the content from path ${path}.`);
+        }
+      }
+    } catch {
+      logger.debug({ path }, 'GitHub: Content not found');
     }
+    return null;
```
l0b0, line 111: Don't we want to throw the exception all the way out if there's a problem? Then we won't have to special-case `null` handling in callers.

src/utils/github.ts
```diff
@@ -94,15 +94,21 @@ export class GithubApi {
...
     logger.info({ path }, 'GitHub API: Get Content');
-    const response = await this.octokit.rest.repos.getContent({ owner: this.owner, repo: this.repo, path });
-    if (!this.isOk(response.status)) throw new Error('Failed to get aerial TileSet config.');
-    if ('content' in response.data) {
-      return Buffer.from(response.data.content, 'base64').toString();
-    } else {
-      throw new Error(`Unable to find the content from path ${path}.`);
+    try {
+      const response = await this.octokit.rest.repos.getContent({ owner: this.owner, repo: this.repo, path });
+      if (this.isOk(response.status)) {
```
l0b0, line 101: There's no `else` clause here, so failure responses are not handled or reported. Is this intentional?

src/utils/github.ts
```diff
@@ -94,15 +94,21 @@ export class GithubApi {
...
+        if ('content' in response.data) {
```
l0b0, line 102: Is it possible for `response.data.content` not to exist if the response is OK?

src/utils/github.ts
```diff
@@ -94,15 +94,21 @@ export class GithubApi {
...
+          return Buffer.from(response.data.content, 'base64').toString();
+        } else {
+          throw new Error(`Unable to find the content from path ${path}.`);
+        }
+      }
+    } catch {
+      logger.debug({ path }, 'GitHub: Content not found');
```
l0b0, line 109: Why throw an exception and catch it again immediately?

src/commands/basemaps-github/make.cog.github.ts
```diff
@@ -160,6 +169,96 @@ export class MakeCogGithub {
...
+      const featureChange = Number(newLayer['lds:feature_count']) - Number(existingLayer['lds:feature_count']);
+
+      if (newLayer['lds:version'] === existingLayer['lds:version'] && featureChange !== 0) {
+        // Alert if feature changed with no version bump.
+        return `🟥🟥🟥🟥 Feature Change Detected ${newLayer['lds:name']} - version: ${newLayer['lds:version']} features: ${newLayer['lds:feature_count']} (+${featureChange}) 🟥🟥🟥🟥`;
+      }
+
+      if (featureChange >= 0) {
+        // Add Features
+        return `🟦 ${newLayer['lds:name']} - version: ${newLayer['lds:version']} (from: ${existingLayer['lds:version']}) features: ${newLayer['lds:feature_count']} (+${featureChange})`;
```
l0b0, line 187: What if `featureChange === 0`? I would assume that would return null.

src/commands/basemaps-github/make.cog.github.ts
```diff
@@ -160,6 +169,96 @@ export class MakeCogGithub {
...
+    if (newCollection == null) throw new Error(`Failed to get target collection json from ${newCollectionPath}.`);
+    const ldsLayers = newCollection.links.filter((f) => f.rel === 'lds:layer') as StacLinkLds[];
+
+    // Log all the new inserts for new tileset
+    if (existingTileSet == null) {
+      changes.push(`New TileSet ts_${layer.name} with layer ${layer.name}.\n`);
+      for (const l of ldsLayers) {
+        changes.push(this.getVectorChanges(l, undefined));
+      }
+      return changes.filter((f) => f != null).join('\n');
```
l0b0, line 226: Instead of filtering here we could push only non-null items on line 224.

src/commands/basemaps-github/make.cog.github.ts
```diff
@@ -160,6 +169,96 @@ export class MakeCogGithub {
...
+    }
+
+    // Compare the different of existing tileset, we usually only have one layers in the vector tiles, so the loop won't fetch very much
+    for (const l of existingTileSet.layers) {
+      if (l[3857] == null) continue;
```
l0b0, line 231: This magic value should probably be a constant.

src/commands/basemaps-github/make.cog.github.ts
```diff
@@ -160,6 +169,96 @@ export class MakeCogGithub {
...
+      if (l.name !== layer.name) continue;
```
l0b0, line 232: We could avoid both this and the above line by looping over something like `existingTileSet.layers.filter((existingLayer) => {existingLayer[3857] != null && existingLayer.name === layer.name})`.

src/commands/basemaps-github/make.cog.github.ts
Wentao-Kuang: The lds layer might be upgrade version with not new features, maybe just date update on existing feature. In contrast, we shouldn't have a feature change if no version upgrade.

src/utils/github.ts
Wentao-Kuang: This is not working as the content may not inthe data type.

src/utils/github.ts
Wentao-Kuang: The response status always return 200 unless catch an error from `this.octokit.rest.repos.getContent`, we might get 404 status from the error. This octokit Api sucks.

src/commands/basemaps-github/make.cog.github.ts
Wentao-Kuang: I think the filter will add another loop? We already got the checks and continues above which make the `existingTileSet.layers` only loop once.

src/commands/basemaps-github/make.cog.github.ts
Wentao-Kuang: I think that's okey, as we already defined the type `changes: (string | null)[]`, so we might be better to filter at end.

src/utils/github.ts
Wentao-Kuang: We need the above error if the content not exist in data.

src/utils/github.ts
Wentao-Kuang: The way I am trying to handling is to not error out the code from failure response here, and only return null if any errors, so we treat all error as no content found from the github.

src/commands/basemaps-github/create-pr.ts
Wentao-Kuang, line 19:
```suggestion
  if (!validBuckets.has(bucket)) {
```

src/utils/github.ts
l0b0: Returning `null` in case of any error means we lose information, which means we can't debug without changing the code.

src/utils/github.ts
l0b0: In that case the `if` statement can just be removed, because the code is unreachable if it's not OK.

src/commands/basemaps-github/make.cog.github.ts
l0b0: That assumption should probably be part of the code.

Wentao-Kuang: That's the line 182 doing this check first.

src/commands/basemaps-github/make.cog.github.ts
l0b0: The point is the code will be faster and simpler if we filter immediately rather than at the end.

src/commands/basemaps-github/make.cog.github.ts
l0b0: You mean line 181? That has *two* clauses, so it's entirely possible for `featureChange` to be zero once it gets to this line.

src/commands/basemaps-github/make.cog.github.ts
Wentao-Kuang: I mean, because the array allows `null`, we might accidentally adding new logic or somewhere that adds null in the array next time? It seems safer to remove this at the end. Also the maximum changes for our vector maps is about 400 lds layers. No harm to filter again here.

src/utils/github.ts
Wentao-Kuang: I have updated this to return null only when a 404 status been catched.

l0b0: But now the code ignores all other errors. You probably want a `throw error` on line 116.

src/utils/github.ts
l0b0: I mean, why not log it immediately, rather than throwing and catching?

Wentao-Kuang: Good point, just added that.

src/utils/github.ts
Wentao-Kuang: Good point, just added that.

src/utils/github.ts
Wentao-Kuang: I think this should be error if response content not exists. If we log and return null. This error won't been capture.

src/commands/basemaps-github/make.cog.github.ts
blacha: if you hover over `l` this shows the type as a `EpsgCode`.

The EpsgCode `3857` in our code base is generally just `3857`, the naming of it in some places is shown as `Epsg.WebMercator` or `Epsg.Google` but these both make it much harder to understand the actual code and then you need more knowledge of our system to know that `Epsg.Google` is referring back to 3857.

l0b0 (dismissed):

l0b0 (approved):

src/utils/github.ts
```diff
@@ -94,15 +94,17 @@ export class GithubApi {
...
-    if (!this.isOk(response.status)) throw new Error('Failed to get aerial TileSet config.');
-    if ('content' in response.data) {
-      return Buffer.from(response.data.content, 'base64').toString();
-    } else {
-      throw new Error(`Unable to find the content from path ${path}.`);
-    }
+    const response = await this.octokit.rest.repos
+      .getContent({ owner: this.owner, repo: this.repo, path })
+      .catch((e) => {
+        if (e.status === 404) return null;
```
blacha, line 102:
```suggestion
        if (e?.status === 404) return null;
```

in theory e could be null here, very unlikely but could happen.

l0b0 (approved):

blacha (approved):

PR #946, opened 2024-04-01
fix: Use Pacific/Auckland time zone for collection years TDE-1066

l0b0:
#### Motivation

The original code used the original UTC datetimes in the STAC collection to work out the start and end years. But we want to use the local start and end years.

#### Modification

Localise datetime string before extracting the year.

#### Checklist

- [x] Tests updated
- [ ] Docs updated (N/A)
- [x] Issue linked in Title

### DIFF
--- a/src/commands/path/path.generate.ts
+++ b/src/commands/path/path.generate.ts
@@ -131,8 +131,8 @@ export function formatName(region: string, geographicDescription?: string): stri
 
 export function formatDate(collection: StacCollection): string {
   const interval = collection.extent?.temporal?.interval?.[0];
-  const startYear = interval[0]?.slice(0, 4);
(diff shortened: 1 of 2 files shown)


amfage (dismissed):

src/commands/path/path.generate.ts
```diff
@@ -143,6 +143,21 @@ export function formatDate(collection: StacCollection): string {
   return `${startYear}-${endYear}`;
 }
 
+function getPacificAucklandYear(dateString: string | null): string | undefined {
```
paulfouquet, line 146: Would be good to have the description of this function as JSDoc

paulfouquet (approved):

amfage (approved):

PR #949, opened 2024-04-03
feat: EAI_AGAIN middleware TDE-1114

MDavidson17:
#### Motivation

Retry workflows that fail due to EAI_AGAIN errors. 

#### Modification

Added an SDK middleware to try 3 times when receiving an `EAI_AGAIN` error. 

#### Checklist

- [x] Tests updated
- [ ] Docs updated -n/a
- [x] Issue linked in Title

### DIFF
--- a/src/__test__/eai_againRetry.test.ts
+++ b/src/__test__/eai_againRetry.test.ts
@@ -26,19 +26,19 @@ describe('eai_againRetryMiddleware', () => {
   });
   it('should run next once if it succeeds', () => {
     const fakeNext = fakeNextBuilder(0);
-    eaiAgainBuilder((attempt: number) => attempt * 0)(fakeNext, {})({ input: {}, request: {} });
+    eaiAgainBuilder(() => 0)(fakeNext, {})({ input: {}, request: {} });
     assert.equal(callCount, 1);
   });
 
   it('should try three times when getting EAI_AGAIN errors', async () => {
     const fakeNext = fakeNextBuilder(2);
-    await eaiAgainBuilder((attempt: number) => attempt * 0)(fakeNext, {})({ input: {}, request: {} });
+    await eaiAgainBuilder(() => 0)(fakeNext, {})({ input: {}, request: {} });
     assert.equal(callCount, 3);
   });
 
   it('should throw error if next fails three times', () => {
     const fakeNext = fakeNextBuilder(3);
-    assert.rejects(eaiAgainBuilder((attempt: number) => attempt * 0)(fakeNext, {})({ input: {}, request: {} }), {
+    assert.rejects(eaiAgainBuilder(() => 0)(fakeNext, {})({ input: {}, request: {} }), {
       message: 'EAI_AGAIN maximum tries (3) exceeded',
     });
   });
--- a/src/__test__/eai_againRetry.test.ts
+++ b/src/__test__/eai_againRetry.test.ts
@@ -3,7 +3,7 @@ import { beforeEach, describe, it } from 'node:test';
 
 import { BuildHandler, MetadataBearer } from '@smithy/types';
 
-import { eaiAgain } from '../fs.register.js';
+import { eaiAgainBuilder } from '../fs.register.js';
 
 let callCount = 0;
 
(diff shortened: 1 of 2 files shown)


src/fs.register.ts
```diff
@@ -31,9 +32,34 @@ export const fqdn: FinalizeRequestMiddleware<object, MetadataBearer> = (next) =>
...
+    const maxTries = 3;
+    return async (args) => {
+      for (let i = 0; i < maxTries; i++) {
+        try {
+          return await next(args);
+        } catch (error) {
+          if (error && typeof error === 'object' && 'code' in error && error.code !== 'EAI_AGAIN') {
+            throw error;
+          }
+          logger.error('eai_again:retry:' + i);
```
blacha, line 49: what other information could we log here? eg hostname? would be interesting to see if we are getting the retries on specific hosts only.

I also think this is maybe just a `warn` and `error` when the retries limit is hit.

src/fs.register.ts
```diff
@@ -31,9 +32,34 @@ export const fqdn: FinalizeRequestMiddleware<object, MetadataBearer> = (next) =>
...
+      throw new Error(`EAI_AGAIN maximum tries (${maxTries}) exceeded`);
+    };
+  };
+  return eaiAgain;
+}
+
 const client = new S3Client();
 export const s3Fs = new FsAwsS3V3(client);
 client.middlewareStack.add(fqdn, { name: 'FQDN', step: 'finalizeRequest' });
+client.middlewareStack.add(eaiAgainBuilder(1000), { name: 'EAI_AGAIN', step: 'build' });
```
blacha, line 62: When seeing a retry handler I generally assume the frist number is the number of retries not the backoff, it is also a good idea to backoff further as the retry count expands.

the S3Client retry handler is configured :

```typescript
new ConfiguredRetryStrategy(
  4, // Retry count
  (attempt: number) => 100 + attempt * 1000 // backoff function.
)
```

could we just copy this interface?

src/__test__/eai_againRetry.test.ts
```diff
@@ -0,0 +1,45 @@
...
+  return fakeNext;
+}
+
+describe('eai_againRetryMiddleware', () => {
+  beforeEach(() => {
+    callCount = 0;
+  });
+  it('should run next once if it succeeds', () => {
+    const fakeNext = fakeNextBuilder(0);
+    eaiAgainBuilder(0)(fakeNext, {})({ input: {}, request: {} });
```
l0b0, line 29: Is the `, {}` after `fakeNext` actually necessary?

src/fs.register.ts
l0b0, line 62: The `build` step is before `finalizeRequest`, so we should probably swap these two lines.

src/fs.register.ts
MDavidson17: Is there a way to get `error?.hostname` without having to add the very long if statement like in L46 so I don't get this error: `Property 'hostname' does not exist on type '{}'`

l0b0 (approved):

src/fs.register.ts
```diff
@@ -31,8 +32,40 @@ export const fqdn: FinalizeRequestMiddleware<object, MetadataBearer> = (next) =>
...
+      throw new Error(`EAI_AGAIN maximum tries (${maxTries}) exceeded`);
+    };
+  };
+  return eaiAgain;
+}
+
 const client = new S3Client();
 export const s3Fs = new FsAwsS3V3(client);
+client.middlewareStack.add(
+  eaiAgainBuilder((attempt: number) => attempt * 1000),
```
blacha, line 66: `0 * 1000 = 0` which is a instant retry and possibly a instant failure, we should set a minimum delay

```suggestion
  eaiAgainBuilder((attempt: number) => 100 + attempt * 1000),
```

src/fs.register.ts
```diff
@@ -31,8 +32,40 @@ export const fqdn: FinalizeRequestMiddleware<object, MetadataBearer> = (next) =>
...
+    return async (args) => {
+      for (let attempt = 1; attempt <= maxTries; attempt++) {
+        try {
+          return await next(args);
+        } catch (error) {
+          if (error && typeof error === 'object' && 'code' in error && 'hostname' in error) {
+            if (error.code !== 'EAI_AGAIN') {
+              throw error;
+            }
+            logger.warn({ host: error.hostname }, `eai_again retry #${attempt}`);
```
blacha, line 50: its best to structure log the attempt counter, of data so it can be filtered on later.

Could we also add totalDelay? and log the next dealy? 

```suggestion
            const delay = timeout(attempt);
            logger.warn({ host: error.hostname, attempt, delay, totalDelay }, `eai_again:retry`);
            totalDelay += delay;
```

l0b0 (approved):

blacha (approved):

PR #950, opened 2024-04-03
feat(copy): check etag before skipping writes TDE-1172

blacha:
#### Motivation

When copying tiff files, sometimes the header information can change without affecting the file size.

There needs to be a safer way to determine if the files are the same so they can be updated.

ETag is not perfect, as it depending on how the file is uploaded the ETag can vary, but since its the same process doingt he copy

#### Modification

If the file exists and is a `--no-clobber` operation validate that the ETags match.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title


blacha:
Fixed with #1001

PR #953, opened 2024-04-04
feat(basemaps): Update the create-pr cli to support elevation config. BM-936

Wentao-Kuang:
#### Motivation

Upgrade the create-pr cli to support elevation config type.

#### Modification

Breaking change: 
- remove the `--vector` flag in cli and use `--config-type` to select for `raster/elevation/vector` Configs
- `--individual` flag is using to create individual configs for all config types now.
- new config location for individual raster/elevation config. i.e. `config/tileset/auckland/elevation/` or `config/tileset/auckland/imagery`
 
#### Checklist

- [ ] Tests updated - no test
- [ ] Docs updated - no doc
- [x] Issue linked in Title


src/commands/basemaps-github/make.cog.github.ts
```diff
@@ -164,28 +221,43 @@ export class MakeCogGithub {
...
+
+      // skip pull request tileset prepare failure.
+      if (newTileSet == null) throw new Error('Failed to prepare new Vector tileSet.');
+
+      // Github
+      const title = `config(vector): Update the ${this.imagery} to ${filename} config file.`;
+      const content = await prettyPrint(JSON.stringify(newTileSet, null, 2), ConfigPrettierFormat);
+      const file = { path: tileSetPath, content };
+      // Github create pull request
+      await gh.createPullRequest(branch, title, botEmail, [file]);
```
l0b0, line 259: These lines seem to be identical in both branches, so should probably be outside the `if` block.

l0b0 (approved):

l0b0 (commented):
Wrong button.

src/commands/basemaps-github/create-pr.ts
```diff
@@ -104,18 +110,18 @@ export const CommandCreatePRArgs = {
     defaultValue: () => 'linz/basemaps-config',
     defaultValueIsSerializable: true,
   }),
+  configType: option({
+    type: optional(oneOf(Object.values(ConfigType))),
+    long: 'config-type',
+    description: `Basemaps config file type, includes ${[...Object.values(Category)].join(', ')}`,
```
blacha, line 116: Validating against ConfigType but docs are using Category, what is the difference?

Wentao-Kuang: Very good pick. I just copied from other one and forgot to update the type. Just fixed.

src/commands/basemaps-github/make.cog.github.ts
Wentao-Kuang: This main differences here are the `[file]` which containers special `path` and `content` for the pull requests. So it will need be inside the if else block for each case. Otherwise, we might need make a `let` variable for the `file` which is not expected coding convention.

l0b0: Lines 260 through 267 are *identical* to 277 through 284, so they absolutely can be outside the `if` block.

Wentao-Kuang: Yeah, but line 264 requires the `newTileSet` output from 259 and line 265 require `tileSetPath` from 258. Same as the else block. Unless we defined a `let tileSetPath;` and `let newTileSet;` outside the blocks forehead. But we should avoid let variables in typescript, then use const instead.

Another options, is just wrap these four line into a separated function. So we don't need see it repeated.

l0b0: I don't think we should avoid `let` if it means duplicating code.

l0b0 (dismissed):

src/commands/basemaps-github/make.cog.github.ts
```diff
@@ -88,38 +91,81 @@ export class MakeCogGithub {
...
+    const gh = new GithubApi(this.repository);
+    const branch = `feat/bot-config-elevation-${this.imagery}${this.ticketBranchSuffix}`;
+    const title = `config(elevation): Add elevation ${this.imagery} to ${filename} config file.`;
+
+    // Clone the basemaps-config repo and checkout branch
+    logger.info({ imagery: this.imagery }, 'GitHub: Get the master TileSet config file');
+    if (individual) {
+      if (region == null) region = 'individual';
+      // Prepare new standalone tileset config
+      layer.category = category;
```
blacha, line 143: this modifies the source `layer` object which was passed into this function, in general we should avoid updating objects when passing them down into functions unless we are explicit that the function is designed to add to the source bojects.

`const targetLayer = {...layer, minZoom: 0, maxZoom:32 }`

Wentao-Kuang: Just fixed, thank you.

blacha (approved):

l0b0 (approved):

PR #970, opened 2024-04-29
feat(basemaps): Get vector title from the collection.json file. BM-1017

Wentao-Kuang:
#### Motivation
We are populating title inside collection.json file from the etl. The `collection.title` is the combined title when processing multiple vector layers which is using the parameter from `etl --title`. And all the `collection.links` should contain the `lds:title` from the lds export. We should using this title when there is only one lds layer exists in the etl output.

#### Modification
Get the title from collection.json and get the title from `lds:title` for individual vector output.

#### Checklist

- [ ] Tests updated - no test
- [ ] Docs updated - no doc
- [x] Issue linked in Title

### DIFF
--- a/src/commands/basemaps-github/create-pr.ts
+++ b/src/commands/basemaps-github/create-pr.ts
@@ -80,9 +80,20 @@ async function parseVectorTargetInfo(target: string): Promise<{ name: string; ti
 
   if (epsg == null || name == null) throw new Error(`Invalid target ${target} to parse the epsg and imagery name.`);
   if (epsg !== Epsg.Google) throw new Error(`Unsupported epsg code ${epsg.code} for vector map.`);
-  // Try to get the region for individual layers
+  // Try to get the title
+  const collectionPath = fsa.join(target, 'collection.json');
+  const collection = await fsa.readJson<StacCollection>(collectionPath);
+  if (collection == null) throw new Error(`Failed to get target collection json from ${collectionPath}.`);
+  const ldsLayers = collection.links.filter((f) => f.rel === 'lds:layer');
+  let title = collection.title;
+  // Get title from lds:title for individual vector layer
+  if (ldsLayers.length === 1) {
+    const ldsTitle = ldsLayers[0]?.['lds:title'];
+    if (ldsTitle != null) title = String(ldsTitle);
+  }
+  if (title == null) throw new Error(`Failed to get title from collection.json.`);
 
-  return { name: name, epsg: epsg.code, title: name };
+  return { name: name, epsg: epsg.code, title };
 }
 
 export const CommandCreatePRArgs = {


blacha (approved):

src/commands/basemaps-github/create-pr.ts
```diff
@@ -80,9 +80,20 @@ async function parseVectorTargetInfo(target: string): Promise<{ name: string; ti
...
   if (epsg !== Epsg.Google) throw new Error(`Unsupported epsg code ${epsg.code} for vector map.`);
-  // Try to get the region for individual layers
+  // Try to get the title
+  const collectionPath = fsa.join(target, 'collection.json');
+  const collection = await fsa.readJson<StacCollection>(collectionPath);
+  if (collection == null) throw new Error(`Failed to get target collection json from ${collectionPath}.`);
+  const ldsLayers = collection.links.filter((f) => f.rel === 'lds:layer');
+  let title = collection.title;
+  // Get title from lds:title for individual vector layer
+  if (ldsLayers.length === 1) {
```
l0b0, line 90: Why only if there's exactly one layer? Would it be wrong to get the title from the first layer if there's more than one layer?

src/commands/basemaps-github/create-pr.ts
```diff
@@ -80,9 +80,20 @@ async function parseVectorTargetInfo(target: string): Promise<{ name: string; ti
...
+    const ldsTitle = ldsLayers[0]?.['lds:title'];
```
l0b0, line 91: I don't think `?` here is necessary, because the previous line ensures that index 0 exists. Unless you could have a layer which is literally `undefined`, which would probably be a bug.

src/commands/basemaps-github/create-pr.ts
Wentao-Kuang: We should only get the lds title for individual layer. Otherwise, for the vector map that have a mixed of multiple lds layers, we can use the `collection.title`.

src/commands/basemaps-github/create-pr.ts
Wentao-Kuang: I was trying to remove ? this earlier, but the type checker doesn't allows to. Seems the filter returns a `unknown` type from the filter(line 87), which can be `undifned` type as an `unknown`.

src/commands/basemaps-github/create-pr.ts
l0b0: OIC, if there's exactly one layer its title is preferred over the collection title. Strange, but sure.

src/commands/basemaps-github/create-pr.ts
l0b0: I think the standard solution for that would be to declare the type of `ldsLayers`. But since that's more verbose, let's just go with this for now.

l0b0 (approved):

PR #972, opened 2024-04-29
feat: stac validate links checksum TDE-1134

paulfouquet:
#### Motivation

We need a way to validate all the STAC files quickly. Validating the `assets` takes a long time but validating just the JSON files (`links`) should be fast.

#### Modification

- Add the ability to the `stac validate` command to validate the link checksums
- Allow the `stac validate` command to validate only the `link` checksums (using `--checksum-links`) or both `asset` and `links` (`--checksum`)

#### Checklist

- [x] Tests updated
- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,7 +258,7 @@ stac validate s3://linz-imagery-staging/test/stac-validate/collection.json
 stac validate --checksum --recursive s3://linz-imagery-staging/test/stac-validate/collection.json
 ```
 
-- Validate a the `file:checksum` of all links (only links, not assets) inside of a collection:
+- Validate the `file:checksum` of all links (only links, not assets) inside of a collection:
 
 ```bash
 stac validate --checksum-links --recursive s3://linz-imagery-staging/test/stac-validate/collection.json
--- a/README.md
+++ b/README.md
@@ -258,6 +258,12 @@ stac validate s3://linz-imagery-staging/test/stac-validate/collection.json
 stac validate --checksum --recursive s3://linz-imagery-staging/test/stac-validate/collection.json
 ```
 
+- Validate a the `file:checksum` of all links (only links, not assets) inside of a collection:
+
+```bash
+stac validate --checksum-links --recursive s3://linz-imagery-staging/test/stac-validate/collection.json
+```
+
 ### `tileindex-validate`
 
 Validate or create retiling information for a list of tiffs.
--- a/src/commands/stac-catalog/stac.catalog.ts
+++ b/src/commands/stac-catalog/stac.catalog.ts
@@ -1,12 +1,12 @@
 import { fsa } from '@chunkd/fs';
 import { command, option, positional, string } from 'cmd-ts';
-import { createHash } from 'crypto';
 import { isAbsolute } from 'path';
 import * as st from 'stac-ts';
 
 import { CliInfo } from '../../cli.info.js';
 import { logger } from '../../log.js';
-import { config, registerCli, Sha256Prefix, verbose } from '../common.js';
+import { hashString } from '../../utils/hash.js';
+import { config, registerCli, verbose } from '../common.js';
 
 /** is a path a URL */
 export function isUrl(path: string): boolean {
@@ -87,8 +87,7 @@ export async function createLinks(basePath: string, templateLinks: st.StacLink[]
       const relPath = makeRelative(basePath, coll);
       const buf = await fsa.read(coll);
       const collection = JSON.parse(buf.toString()) as st.StacCollection;
-      // Multihash header 0x12 - Sha256 0x20 - 32 bits of hex digest
-      const checksum = Sha256Prefix + createHash('sha256').update(buf).digest('hex');
+      const checksum = hashString(buf);
       const collLink: st.StacLink = {
         rel: 'child',
         href: fsa.join('./', relPath),
--- a/src/commands/stac-catalog/stac.catalog.ts
+++ b/src/commands/stac-catalog/stac.catalog.ts
@@ -5,7 +5,7 @@ import * as st from 'stac-ts';
 
 import { CliInfo } from '../../cli.info.js';
 import { logger } from '../../log.js';
-import { hashString } from '../../utils/hash.js';
+import { hashBuffer } from '../../utils/hash.js';
 import { config, registerCli, verbose } from '../common.js';
 
 /** is a path a URL */
@@ -87,7 +87,7 @@ export async function createLinks(basePath: string, templateLinks: st.StacLink[]
       const relPath = makeRelative(basePath, coll);
       const buf = await fsa.read(coll);
       const collection = JSON.parse(buf.toString()) as st.StacCollection;
-      const checksum = hashString(buf);
+      const checksum = hashBuffer(buf);
       const collLink: st.StacLink = {
         rel: 'child',
         href: fsa.join('./', relPath),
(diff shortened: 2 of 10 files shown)


src/commands/stac-validate/__test__/stac.validate.test.ts
```diff
@@ -60,4 +72,80 @@ describe('stacValidate', function () {
     );
     assert.equal(normaliseHref('./item.json', 'collection.json'), 'item.json');
   });
+
+  it('validateStacChecksum', async () => {
+    const memory = new FsMemory();
+    fsa.register('memory://', memory);
+    const path = 'memory://stac/';
```
l0b0, line 79: All these tests seem to have this as their first lines and `memory.files.clear();` as their last line, so it would probably make sense to pull them out into `beforeEach`/`afterEach` functions. It might be necessary to pull the tests into a separate `describe` to make sure those functions only run for the relevant tests, though.

src/commands/stac-validate/stac.validate.ts
```diff
@@ -213,6 +207,105 @@ export const commandStacValidate = command({
...
+  const checksum: string = stacObject['file:checksum'] as string;
+
+  if (checksum == null) {
+    if (ctx.allowMissing) return true;
+    logger.error({ source, checksum }, 'Validate:Checksum:Missing');
+    return false;
+  }
+  // 12-20 is the starting prefix for all sha256 multihashes
+  if (!checksum.startsWith('1220')) {
+    if (ctx.allowUnknown) return true;
```
l0b0, line 292: Do we want to be able to pass `allowUnknown = true`?

src/commands/stac-validate/stac.validate.ts
```diff
@@ -213,6 +207,105 @@ export const commandStacValidate = command({
...
+ * @returns the list of link paths that failed the validation
+ */
+export async function validateLinks(
+  stacJson: st.StacItem | st.StacCollection | st.StacCatalog,
+  path: string,
+): Promise<string[]> {
+  const linksFailures: string[] = [];
+  for (const link of stacJson.links) {
+    if (link.rel === 'self') continue;
+    // we `allowMissing` because some STAC links might not have a checksum yet (feature added later)
```
l0b0, line 246: I don't understand this comment. Why would any STAC links not have a checksum at this point? Can we enumerate them, so that we can avoid this flag? Also, what's meant by "feature added later"?

src/commands/stac-validate/stac.validate.ts
```diff
@@ -213,6 +207,105 @@ export const commandStacValidate = command({
...
+   */
+  allowUnknown: boolean;
+}
+
+/**
+ * Validate if the checksum found in the stacObject (`file:checksum`) corresponds to its actual file checksum.
+ * @param stacObject
+ * @param path
+ * @param ctx
+ * @returns
```
l0b0, line 274: We should probably either comment on these or remove this part of the comment. As it stands it just duplicates some of the code.

src/commands/stac-validate/stac.validate.ts
```diff
@@ -213,6 +207,105 @@ export const commandStacValidate = command({
...
+  if (source.startsWith('./')) source = fsa.join(dirname(path), source.replace('./', ''));
+  const checksum: string = stacObject['file:checksum'] as string;
+
+  if (checksum == null) {
+    if (ctx.allowMissing) return true;
+    logger.error({ source, checksum }, 'Validate:Checksum:Missing');
+    return false;
+  }
+  // 12-20 is the starting prefix for all sha256 multihashes
+  if (!checksum.startsWith('1220')) {
```
l0b0, line 291: This magic value should probably be a constant somewhere central.

src/commands/stac-validate/stac.validate.ts
paulfouquet: The might not have checksums if the STAC has been generated before the feature that add a checksum for the links. As far as I know, we don't have a process to regenerate the STAC only from an existing dataset.

src/commands/stac-validate/stac.validate.ts
paulfouquet: Probably not at this stage

l0b0 (approved):

src/commands/stac-catalog/stac.catalog.ts
```diff
@@ -88,7 +88,7 @@ export async function createLinks(basePath: string, templateLinks: st.StacLink[]
       const buf = await fsa.read(coll);
       const collection = JSON.parse(buf.toString()) as st.StacCollection;
       // Multihash header 0x12 - Sha256 0x20 - 32 bits of hex digest
-      const checksum = '1220' + createHash('sha256').update(buf).digest('hex');
+      const checksum = Sha256Prefix + createHash('sha256').update(buf).digest('hex');
```
blacha, line 91: it would be better to store extract this as a function as the prefix is only valid for the combination of 'sha256' and 'hex' digest.

there is a function that does this for streams, could put another one next to it for buffers or strings.

README.md
```diff
@@ -258,6 +258,12 @@ stac validate s3://linz-imagery-staging/test/stac-validate/collection.json
 stac validate --checksum --recursive s3://linz-imagery-staging/test/stac-validate/collection.json
 ```
 
+- Validate a the `file:checksum` of all links (only links, not assets) inside of a collection:
```
amfage, line 261:
```suggestion
- Validate the `file:checksum` of all links (only links, not assets) inside of a collection:
```

src/commands/stac-validate/hash.worker.ts
```diff
@@ -13,7 +15,7 @@ export async function hashStream(stream: Readable): Promise<string> {
     stream.on('data', (chunk) => hash.update(chunk));
     // 0x12 - ID of sha256 multi hash
     // 0x20 - 32 bytes (256 bits) of data
```
blacha, line 17: these comments do not make any sense here and should be moved to where the prefix is defined.

src/utils/hash.ts
```diff
@@ -17,3 +21,13 @@ export async function hashStream(stream: Readable): Promise<string> {
...
   });
 }
+
+/**
+ * Create a multihash from a string or Buffer
+ *
+ * @param x a string or `Buffer`
+ * @returns sha256 multihash string of the string or `Buffer`
+ */
+export function hashString(x: Buffer | string): string {
```
l0b0, line 31: It looks like every production use of this takes a `Buffer`, so we should probably do the same. Alternatively, we could copy the [type of the `hash.update` `data` parameter](https://nodejs.org/api/crypto.html#hashupdatedata-inputencoding).

Also, the name `x` for the input should probably be something more readable. This might be one of the rare times when `data` is actually appropriate.

src/utils/hash.ts
```diff
@@ -1,7 +1,11 @@
 import { createHash } from 'crypto';
 import { Readable } from 'stream';
 
-import { Sha256Prefix } from '../common.js';
+/** 1220 is the starting prefix for all sha256 multihashes
+ *  0x12 - ID of sha256 multi hash
+ *  0x20 - 32 bytes (256 bits) of data
+ */
```
l0b0, line 7: It might be useful to link to [the Multihash standard](https://multiformats.io/multihash/) here.

l0b0 (approved):

blacha (approved):

PR #1001, opened 2024-05-31
feat: verify multihash when copy TDE-1181 TDE-1172

paulfouquet:
#### Motivation

When copying files across S3 bucket, we need a way to verify if the file already exists in order to prevent overriding it. The current way to verify that is based on the file size which is not accurate.

#### Modification

Add a verification on the file checksum, stored in its metadata or generated on the go.

#### Checklist

- [x] Tests updated
- [ ] Docs updated (N/A)
- [x] Issue linked in Title


src/commands/copy/__test__/copy.test.ts
```diff
@@ -151,4 +151,140 @@ describe('copyFiles', () => {
...
+          multihash: '12206fd977db9b2afe87a9ceee48432881299a6aaf83d935fbbe83007660287f9c2e',
+          unique: 'fileA',
+        },
+      }),
+      fsa.write('memory://target/topographic.json', Buffer.from(JSON.stringify({ test: true })), {
+        contentType: 'application/octet-stream',
+        metadata: { multihash: 'different', unique: 'fileB' },
+      }),
+    ]);
+    assert.rejects(
```
l0b0, line 169: Can we verify *why* this was rejected?

src/commands/copy/__test__/copy.test.ts
```diff
@@ -151,4 +151,140 @@ describe('copyFiles', () => {
...
+      ],
+      start: 0,
+      size: 1,
+      force: true,
+      noClobber: true,
+      fixContentType: false,
+    });
+    const jsonTarget = await fsa.head('memory://target/topographic.json');
+
+    assert.equal(jsonTarget?.metadata?.['unique'], 'fileB');
```
l0b0, line 220: So we don't copy the file if the contents are the same and the _metadata_ is different? It makes sense in this case, but that seems like something which should be documented.

src/commands/copy/__test__/copy.test.ts
```diff
@@ -151,4 +151,140 @@ describe('copyFiles', () => {
...
+      fsa.write('memory://source/topographic.json', Buffer.from(JSON.stringify({ test: true })), {
+        contentType: 'application/octet-stream',
+        metadata: {
+          multihash: '12206fd977db9b2afe87a9ceee48432881299a6aaf83d935fbbe83007660287f9c2e',
+          unique: 'fileA',
+        },
+      }),
+      fsa.write('memory://target/topographic.json', Buffer.from(JSON.stringify({ test: true })), {
+        contentType: 'application/octet-stream',
+        metadata: { multihash: 'different', unique: 'fileB' },
```
l0b0, line 166: We should probably use a valid multihash. I would expect that the invalid multihash could trigger a condition which cancels the upload for a different reason than the one stated in the test description.

src/commands/copy/__test__/copy.test.ts
```diff
@@ -151,4 +151,140 @@ describe('copyFiles', () => {
...
+  it('should not copy files when different size or multihash but no force', async () => {
+    await Promise.all([
+      fsa.write('memory://source/topographic.json', Buffer.from(JSON.stringify({ test: true })), {
+        contentType: 'application/octet-stream',
+        metadata: {
+          multihash: '12206fd977db9b2afe87a9ceee48432881299a6aaf83d935fbbe83007660287f9c2e',
+          unique: 'fileA',
+        },
+      }),
+      fsa.write('memory://target/topographic.json', Buffer.from(JSON.stringify({ test: true })), {
```
l0b0, line 164: Since this file has a different multihash, it should also have different content for clarity. You could pull out for example `const originalContent = Buffer.from(JSON.stringify({ "a": 1 }))`, `const originalSizeDifferentContent = Buffer.from(JSON.stringify({ "b": 2 }))`, and `const differentSizeContent = Buffer.from(JSON.stringify({ "xxx": 123 }))` and their hashes to use in the various tests.

l0b0 (dismissed):

l0b0 (commented):
Pressed wrong button.

src/commands/copy/__test__/copy.test.ts
paulfouquet: > So we don't copy the file if the contents are the same and the metadata is different?
This isn't true. We copy the file if the content is the same but the `multihash` metadata different for example.

The metadata `unique` is just for testing purpose here and validate that the file has not been copied.

src/commands/copy/__test__/copy.test.ts
l0b0: nixpkgs uses a ["fake" hash](https://github.com/NixOS/nixpkgs/blob/cfbeda848e75c62d6facd3ae40d6f239dcee9375/lib/deprecated/misc.nix#L328-L330) which is visually distinct from any hash you'd actually encounter. We could do the same with something like "1220ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff".

src/commands/copy/__test__/copy.test.ts
l0b0: I understand that, but we might want to document how it works.

amfage (approved):

PR #1010, opened 2024-06-17
fix(basemaps): Fix the individual tileset format to using hex backgroud colour

Wentao-Kuang:
#### Motivation

The last changes update individual config to defined using the ConfigTileSetRaster type which is force the rgb backgroud format. We should using hex backgroud in the individual configs.

#### Modification

Revert back the type to `TileSetConfigSchema` for individual raster tileset. 

#### Checklist

- [ ] Tests updated - No test
- [ ] Docs updated - No doc
- [ ] Issue linked in Title - No ticket


Wentao-Kuang:
We decided to allow rgb in config file. https://github.com/linz/basemaps/pull/3288

PR #1016, opened 2024-06-25
build: upgrade to typescript 5.5

blacha:
#### Motivation

Typescript v5.5 was recently released with many improvements, it would be good to keep upto date with the improved type checking

#### Modification

upgrades to typescript 5.5 and fixes a number of lint rules

Most fixes revolve around `any` and `async` improvements
- casting `.catch(e)` to `unknown` as e is `any`
- casting `JSON.parse` as its return is unknown

Async:
- functions that are async but do not contain a `await`
- promises that are triggered but do not have a `await` or `.catch()`

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/src/commands/stac-github-import/stac.github.import.ts
+++ b/src/commands/stac-github-import/stac.github.import.ts
@@ -10,8 +10,8 @@ import { config, registerCli, verbose } from '../common.js';
 import { prettyPrint } from '../format/pretty.print.js';
 
 const Url: Type<string, URL> = {
-  async from(str) {
-    return new URL(str);
+  from(str) {
+    return Promise.resolve(new URL(str));
   },
 };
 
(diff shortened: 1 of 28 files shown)


src/commands/stac-github-import/stac.github.import.ts
```diff
@@ -11,7 +11,7 @@ import { prettyPrint } from '../format/pretty.print.js';
 
 const Url: Type<string, URL> = {
   async from(str) {
-    return new URL(str);
+    return Promise.resolve(new URL(str));
```
Wentao-Kuang, line 14: Do we still need return `Promise.resolve` for async function?

Wentao-Kuang (approved):

amfage (approved):
:+1:

Wentao-Kuang (approved):

PR #1074, opened 2024-09-20
fix(basemaps): Get the epsg and name correctly for elevation s3 path. BM-1088

Wentao-Kuang:
#### Motivation

Elevation s3 path is one more directory and the imagery s3 path. So the validate path need to get the espg and name with the correct index.

elevation: s3://linz-basemaps/elevation/3857/canterbury_2020-2023_dem_1m/01J868QZZ4RJ485PMWP1ECQYD0
imagery: s3://linz-basemaps/2193/gebco_2020_Nztm2000Quad_305-75m/01F1BFJN8R8P7BXN3XTHC5MT5G/

#### Modification

Get the espg and name from correct index for elevations.

#### Checklist

- [x] Tests updated
- [ ] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/src/commands/basemaps-github/create-pr.ts
+++ b/src/commands/basemaps-github/create-pr.ts
@@ -29,14 +29,15 @@ function assertValidBucket(bucket: string, validBuckets: Set<string>): void {
 
 async function parseRasterTargetInfo(
   target: string,
+  elevation: boolean,
   individual: boolean,
 ): Promise<{ name: string; title: string; epsg: EpsgCode; region: string | undefined }> {
   logger.info({ target }, 'CreatePR: Get the layer information from target');
   const url = new URL(target);
   const bucket = url.hostname;
   const splits = url.pathname.split('/');
-  const epsg = Epsg.tryGet(Number(splits[1]));
-  const name = splits[2];
+  const epsg = elevation ? Epsg.tryGet(Number(splits[2])) : Epsg.tryGet(Number(splits[1]));
+  const name = elevation ? splits[3] : splits[2];
 
   assertValidBucket(bucket, validTargetBuckets);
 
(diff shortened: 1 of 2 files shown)


src/commands/basemaps-github/create-pr.ts
```diff
@@ -29,14 +29,15 @@ function assertValidBucket(bucket: string, validBuckets: Set<string>): void {
...
   individual: boolean,
 ): Promise<{ name: string; title: string; epsg: EpsgCode; region: string | undefined }> {
   logger.info({ target }, 'CreatePR: Get the layer information from target');
   const url = new URL(target);
   const bucket = url.hostname;
   const splits = url.pathname.split('/');
-  const epsg = Epsg.tryGet(Number(splits[1]));
-  const name = splits[2];
+  const epsg = elevation ? Epsg.tryGet(Number(splits[2])) : Epsg.tryGet(Number(splits[1]));
+  const name = elevation ? splits[3] : splits[2];
```
blacha, line 40: parsing the path names are known to be super brittle, could we grab this information from the `collection.json` ?

Wentao-Kuang: Sounds better option, we can validate collection.json first before validate these espg and name

blacha: yeah just grab the collection.json then start looking for this information.

Wentao-Kuang: I didn't seen any name from `collection.json/covering.json/source.json`. Only saw the `tilematrix` in the item.json that we could probably use to parse to epsg.

Wentao-Kuang: @blacha I don't think we can get the name from Stac file atm as we only store title in it.

blacha (approved):

PR #1075, opened 2024-09-20
feat: generate CLI README.md

blacha:
#### Motivation

I do not like having to update the same information in two places

#### Modification

Add script to generate a README.md for all CLIs that are registered in the `AllCommands`
Generate the READMEs on every push to master to validate they have not been modified.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title


src/commands/copy/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Usage
+
+copy <options> [...location]
+
+### Arguments
+
+| Usage         | Description              | Options |
+| ------------- | ------------------------ | ------- |
+| [...location] | Manifest of file to copy |         |
```
l0b0, line 13: `[...foo]` is not a syntax I'm familiar with. What's meant here? Is it zero-or-one locations? If so, `[location]` is more common syntax for that. If it's exactly one location, that would normally be `<location>`.

src/commands/stac-validate/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+| Usage         | Description                            | Options |
+| ------------- | -------------------------------------- | ------- |
+| [...location] | Location of the STAC files to validate |         |
+
+### Options
+
+| Usage                  | Description                            | Options  |
+| ---------------------- | -------------------------------------- | -------- |
+| --config <str>         | Location of role configuration file    | optional |
```
l0b0, line 19: Aren't all options and flags optional?

l0b0 (commented):
Love it!

src/commands/copy/README.md
blacha: this is just the output of the help text, I would imagine it means `string[]` as `...` is the javascript spread operator https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

src/commands/stac-validate/README.md
blacha: there are quite a few that are not optional?

src/commands/copy/README.md
l0b0: So we want to use that syntax? OK.

l0b0 (approved):

PR #1092, opened 2024-10-01
fix: ensure file systems with matching roleArns are registered correctly

blacha:
#### Motivation

file systems are only created for unique roleArns so FileSystemCreated event is not fired twice if two configuration objects have the same roleArn, 

give a confiugration with two buckets both using `roleA` on a service that has a default role `roleDefault`
```
s3://foo/ - roleA
s3://bar/ - roleA
```

requests to 

```typescript
fs.read("s3://foo"); // tries roleDefault then uses roleA
fs.read("s3://foo"); // uses cached roleA
```

depending on the order of reads the default role may be used far too often

```typescript
fs.read("s3://foo") // tries roleDefault then uses roleA
fs.read("s3://bar") // tries roleDefault then uses roleA
fs.read("s3://bar") // tries roleDefault then uses roleA
```

after this change

```typescript
fs.read("s3://foo") // tries roleDefault then uses roleA
fs.read("s3://bar") // tries roleDefault then uses roleA
fs.read("s3://bar") // uses cached roleA
```

#### Modification

hook the file system finder when it needs to find a new file system use that file system to register onto `fsa` 

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/src/__test__/fs.test.ts
+++ b/src/__test__/fs.test.ts
@@ -1,11 +1,12 @@
-import { beforeEach, describe, it } from 'node:test';
+import { after, before, beforeEach, describe, it } from 'node:test';
 
-import { fsa } from '@chunkd/fs';
-import { FsAwsS3V3 } from '@chunkd/source-aws-v3';
+import { FileSystemAbstraction, fsa } from '@chunkd/fs';
+import { S3LikeV3 } from '@chunkd/source-aws-v3';
 import { FsMemory } from '@chunkd/source-memory';
+import { FinalizeRequestMiddleware, MetadataBearer } from '@smithy/types';
 import assert from 'assert';
 
-import { setupS3FileSystem } from '../fs.register.js';
+import { registerFileSystem } from '../fs.register.js';
 
 export class HttpError extends Error {
   statusCode: number;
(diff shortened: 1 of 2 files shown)


src/__test__/fs.test.ts
```diff
@@ -0,0 +1,146 @@
...
+  });
+
+  it('should register on 403', async () => {
+    assert.equal(fsa.systems.length, 1);
+    const s3Fs = registerFileSystem({ config: 'memory://config.json' });
+    s3Fs.client.middlewareStack.add(throw403, throw403Init);
+    assert.deepEqual(fsSystemsPath(), ['s3://']);
+
+    s3Fs.credentials.onFileSystemCreated = (_ac, fs): void => {
+      const fsS3 = fs as FsAwsS3;
```
l0b0, line 113: Do we need this intermediate variable? Could we set the type on the `fs` parameter instead?

src/fs.register.ts
```diff
@@ -63,40 +62,74 @@ export function eaiAgainBuilder(timeout: (attempt: number) => number): BuildMidd
...
+  fsClient.credentials.find = async (path: string): Promise<FsAwsS3 | null> => {
+    const accountConfig = await fsClient.credentials.findCredentials(path);
+    if (accountConfig == null) return null;
+
+    const fileSystem = await oldFind(path);
+    if (fileSystem == null) return null;
+
+    logger.debug({ prefix: path, roleArn: accountConfig.roleArn }, 'FileSystem:Register');
+    fsa.register(accountConfig.prefix, fileSystem);
+    if (fileSystem.s3 && 'client' in fileSystem.s3) {
```
l0b0, line 85: Don't we usually use `fileSystem.s3.client != null` for these kinds of checks?

blacha: When you have a object which you dont know much about you have to do a `"client" in foo` to allow the type checker to narrow the type down.

a minimal example

```typescript
type foo = { bar: 'baz' };

foo.client == null; // Error

'client' in foo && foo.client == null; // Ok
```

This could be reworded to `fileSystem.s3 != null && 'client' in fileSystem.s3` to be more specific though.

blacha, line 85:
```suggestion
    if (fileSystem.s3 != null && 'client' in fileSystem.s3) {
```

src/__test__/fs.test.ts
blacha: We need it for the type casting, setting it on the function would change the function signature causing a typing error as `onFilesystemCreated` needs a specific function signature.

paulfouquet (approved):

src/__test__/fs.test.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    await s3Fs.credentials.find('s3://_linz-topographic/foo.json');
+    await s3Fs.credentials.find('s3://_linz-topographic-upload/foo.json');
+
+    const fileSystems = [...s3Fs.credentials.fileSystems.values()];
+    assert.equal(fileSystems.length, 1);
+    const newFs = fileSystems[0]!.s3 as S3LikeV3;
+
+    assert.equal(
+      newFs.client.middlewareStack.identify().find((f) => f.startsWith('FQDN -')),
```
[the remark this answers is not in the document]
blacha: the AWS SDK forces the middleware to be unique by name, so using `find` or `filter` are somewhat interchangeable here. its proabley more correct to use a `filter`

I generally prefer not to catch errors from dependencies in nodejs especially in something like the aws sdk, where sometime in the future they may change the wording of the error causing us to not catch it any more, or worse change the behaviour to something else.

src/__test__/fs.test.ts
```diff
@@ -0,0 +1,146 @@
...
+      'FQDN - finalizeRequest',
+    );
```
blacha, line 103:
```suggestion
    assert.deepEqual(
      newFs.client.middlewareStack.identify().filter((f) => f.startsWith('FQDN -')),
      ['FQDN - finalizeRequest'],
    );
```

src/__test__/fs.test.ts
```diff
@@ -0,0 +1,146 @@
...
+  };
+  const throw403Init = {
+    name: 'throw403',
+    step: 'initialize',
+    priority: 'high',
+  } as const;
+
+  /** find all the file systems related to s3:// */
+  const fsSystemsPath = (): string[] => {
+    fsa.get('s3://', 'r'); // ensure systems' array is sorted
```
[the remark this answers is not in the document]
blacha: I did look at a few options here, we could typecast that private function sot its not private then call it, it was easier to just run the `.get()`

the typecasting could be done with something like `(fsa as unknown as { sortSystems(): void }).sortSystems()` you have to go via unknown as I couldnt easily convert the private. this leads to other problems due to the introduction of a unknown.

l0b0 (approved):

paulfouquet (approved):

amfage (approved):

PR #1097, opened 2024-10-03
feat: annotate commands with examples

blacha:
#### Motivation

Storing examples with the code makes it easier to update the examples at the same time the code is updated

#### Modification

Annotate commands with a examples in markdown

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title


paulfouquet (dismissed):

src/commands/stac-sync/stac.sync.ts
```diff
@@ -36,6 +38,8 @@ export const commandStacSync = command({
   },
 });
 
+annotateExample(commandStacSync, 'Sync STAC to s3', md.code('bash', 'stac sync /path/to/stac/ s3://linz-imagery/'));
```
amfage, line 41: Do we want to use `linz-imagery` as an example or `nz-imagery`?

src/readme/markdown.ts
```diff
@@ -0,0 +1,16 @@
+export const md = {
+  /**
+   * Create a markdown code block
+   *
+   * @param body body of the code block
+   * @param lang language to use eg `bash`
+   * @returns markdown formatted code block
+   */
+  code(lang: 'bash' | 'typescript', body: string): string {
```
l0b0, line 9: Why allow only these two languages?

blacha: mostly to prevent the old `code(a,b)` vs `code(b,a)` problem as they would both be strings, other langs can be added as used.

l0b0: Would it be better to use an enum (or enum-like construct) for this?

blacha: nope, enums are considered bad as they are not a standard, string unions are your best bet.

if the string union list gets long or needs to be reused, it could be extracted into a standalone type eg `export type MdCodeType = 'bash' | 'python' | 'typescript' | 'json'` but it is currently neither long or reused.

src/commands/stac-sync/stac.sync.ts
blacha, line 41:
```suggestion
annotateExample(commandStacSync, 'Sync STAC to s3', md.code('bash', 'stac sync /path/to/stac/ s3://nz-imagery/'));
```

src/commands/stac-sync/README.md
```diff
@@ -6,6 +6,14 @@ Sync STAC files
 
 stac-sync <options> <str> <arg>
 
+## Examples
+
+#### Sync STAC to s3
+
+```bash
+stac sync /path/to/stac/ s3://linz-imagery/
```
amfage, line 14: Another `linz-imagery`

blacha: ha yeah! my quick fix didnt regenerate these files :sob: fixed below.

l0b0 (approved):

PR #1108, opened 2024-10-14
feat: Flatten providers into strings per role TDE-1291

l0b0:
#### Motivation

Make this information easier to process for GIS tools.

#### Checklist

- [x] Tests updated
- [ ] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/src/commands/mapsheet-coverage/__test__/mapsheet.coverage.test.ts
+++ b/src/commands/mapsheet-coverage/__test__/mapsheet.coverage.test.ts
@@ -122,7 +122,6 @@ describe('mapsheet-coverage', () => {
       id: 'layer-a-id',
       license: 'CC-BY-4.0',
       source: 'ms://layers/a/collection.json',
-      host: '',
       licensor: 'First provider',
       processor: 'First provider',
       producer: 'First provider, Second provider',
--- a/src/commands/mapsheet-coverage/__test__/mapsheet.coverage.test.ts
+++ b/src/commands/mapsheet-coverage/__test__/mapsheet.coverage.test.ts
@@ -37,6 +37,16 @@ function fakeCollection(id: string): StacCollection {
     description: `Layer ${id.toUpperCase()} Description`,
     assets: { capture_area: { href: './capture-area.json' } },
     links: [],
+    providers: [
+      {
+        name: 'First provider',
+        roles: ['producer', 'licensor', 'processor'],
+      },
+      {
+        name: 'Second provider',
+        roles: ['producer'],
+      },
+    ],
   } as unknown as StacCollection;
 }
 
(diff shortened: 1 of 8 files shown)


src/commands/mapsheet-coverage/__test__/mapsheet.coverage.test.ts
```diff
@@ -112,6 +122,10 @@ describe('mapsheet-coverage', () => {
       id: 'layer-a-id',
       license: 'CC-BY-4.0',
       source: 'ms://layers/a/collection.json',
+      host: '',
```
blacha, line 125: I would prefer we don't set things to "" if it doesn't exist it should just be undefined.

src/commands/mapsheet-coverage/mapsheet.coverage.ts
```diff
@@ -144,7 +145,26 @@ export const commandMapSheetCoverage = command({
       captureArea.properties['description'] = collection.description;
       captureArea.properties['id'] = collection.id;
       captureArea.properties['license'] = collection.license;
-      captureArea.properties['providers'] = collection.providers;
+      if (collection.providers) {
+        /*
+         We can't loop over `StacRoles from 'stac-ts'`, so we need to keep this list in sync with that type
```
blacha, line 150: we can reuse the type

```
const Roles: Record<StacRoles, string[]> = { producer: [], processor: [], licensor: [], host: []}
```

This style will lead to runtime errors though as the STAC could contain other roles (eg invalid or newer stac versions), for instance `providers: [ { roles: ['abc123'], name: 'foo' } ]` which would cause a runtime error here.

I would recommend using a set so we de-dupe providers and allow unknown roles?

```typescript
const providerByRoles = new Map<string, Set<string>>();
for (const f of collection.providers ?? []) {
  for (const role of f.roles ?? []) {
    const roleSet = providerByRoles.get(role) ?? new Set();
    roleSet.add(f.name);
    providerByRoles.set(role, roleSet);
  }
}
for (const [name, values] of providerByRoles) captureArea.properties[name] = [...values].join(', ');
```

src/commands/mapsheet-coverage/__test__/mapsheet.coverage.test.ts
l0b0: I like including it, since it means users will know that it's empty, rather than not knowing that it exists. But I'd absolutely leave this up to you guys. @dwsilk?

paulfouquet:
NIT: we don't have a convention in place in this repo for specifying the domain of the commit, example feat(**GeoJSON**)

l0b0:
> NIT: we don't have a convention in place in this repo for specifying the domain of the commit, example feat(**GeoJSON**)

Not sure whether we need a convention; isn't the domain meant to be more of a free-form hint about what the subject is?

blacha:
from the conventional commits spec:

> A scope MAY be provided after a type. A scope MUST consist of a noun describing a section of the codebase surrounded by parenthesis, e.g., fix(parser)

Im not sure we use many scopes in this repo, In the commit history the command name is sometimes used as the scope eg `feat(copy)` 

I am not sure we have a "geojson" section of the codebase so IMO this should be either no scope or related to the command "mapsheet-coverage" ? 

It would be nice to document the common scopes if we do want to be consistent the linter can limit which scopes we use.

src/commands/mapsheet-coverage/__test__/mapsheet.coverage.test.ts
dwsilk: Let's just leave it out completely, I'm curious to see what the end result will look like in GIS software.

blacha (approved):

PR #1113, opened 2024-10-15
build: allow pushing pull request builds to aws ecr

blacha:
#### Motivation

Most of our argo templates only allow tags to be changed a lot of them use our ECR based containers, so pull reuqest builds cannot be used inside of argo.

#### Modification

push to AWS ECR if the CI role exists.

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/.github/workflows/pull-request-container.yml
+++ b/.github/workflows/pull-request-container.yml
@@ -15,6 +15,9 @@ jobs:
     environment:
       name: nonprod
 
+    env:
+      AWS_CI_ROLE: ${{ secrets.AWS_CI_ROLE }}
+
     permissions:
       id-token: write
       contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request-container.yml
```diff
@@ -56,6 +59,19 @@ jobs:
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Configure AWS Credentials
+        if: ${{ env.AWS_CI_ROLE != '' }}
```
l0b0, line 63: Is this check necessary? It's unconditionally included above, and secrets aren't scoped, so the only way this could happen is if we set it to the empty string (or possibly if we remove the secret).

blacha: I believe it was a protection for forks, so if you fork the repository the builds don't fail.

l0b0 (approved):

PR #1136, opened 2024-11-18
feat: generate path from slug TDE-1319

amfage:
#### Motivation

The previous logic to construct the ODR path has partially been replaced by the `linz:slug` field that is constructed in the [stac-setup Argo Tasks command](https://github.com/linz/argo-tasks/tree/master/src/commands/stac-setup). `generate-path` can be simplified to use the pre-constructed slug when publishing to the ODR.

#### Modification

- Read the `linz:slug` field from the standardised collection.json file and use this to construct the ODR path.
- Remove unused functions from the `generate-path` command.
- Refactor where it makes sense to reduce duplication of code.

#### Checklist

- [x] Tests updated
- [x] Docs updated
- [x] Issue linked in Title


src/commands/common.ts
```diff
@@ -159,3 +159,23 @@ export const UrlFolder: Type<string, URL> = {
...
+  'linz:lifecycle': string;
+  'linz:geospatial_category': string;
+  'linz:region': string;
+  'linz:slug': string;
+  'linz:security_classification': string;
+  'linz:event_name'?: string;
+  'linz:geographic_description'?: string;
+}
+
+export const geospatialDataCategories = {
```
paulfouquet, line 173:
```suggestion
export const GeospatialDataCategories = {
```

src/commands/generate-path/path.generate.ts
```diff
@@ -101,48 +79,33 @@ export const commandGeneratePath = command({
...
-
-  if (metadata.category === dataCategories.SCANNED_AERIAL_PHOTOS) {
+  if (metadata.geospatialCategory === geospatialDataCategories.SCANNED_AERIAL_PHOTOS) {
     // nb: Historic Imagery is out of scope as survey number is not yet recorded in collection metadata
-    throw new Error(`Automated target generation not implemented for historic imagery`);
-  }
-
-  if ([dataCategories.URBAN_AERIAL_PHOTOS, dataCategories.RURAL_AERIAL_PHOTOS].includes(metadata.category)) {
-    return `s3://${metadata.targetBucketName}/${metadata.region}/${surveyName}_${metadata.gsd}m/rgb/${metadata.epsg}/`;
+    throw new Error(`Historic Imagery ${metadata.geospatialCategory} is out of scope for automated slug generation.`);
```
paulfouquet, line 84: Is that a typo?

```suggestion
    throw new Error(`Historic Imagery ${metadata.geospatialCategory} is out of scope for automated path generation.`);
```

src/commands/generate-path/path.generate.ts
```diff
@@ -101,48 +79,33 @@ export const commandGeneratePath = command({
...
   }
 
-  if ([dataCategories.DEM, dataCategories.DSM].includes(metadata.category)) {
-    return `s3://${metadata.targetBucketName}/${metadata.region}/${surveyName}/${metadata.category}_${metadata.gsd}m/${metadata.epsg}/`;
+  if ([geospatialDataCategories.DEM, geospatialDataCategories.DSM].includes(metadata.geospatialCategory)) {
+    return `s3://${metadata.targetBucketName}/${metadata.region}/${metadata.slug}/${metadata.geospatialCategory}_${metadata.gsd}m/${metadata.epsg}/`;
   }
 
-  throw new Error(`Path Can't be generated from collection as no matching category: ${metadata.category}.`);
+  throw new Error(`Path Can't be generated from collection as no matching category: ${metadata.geospatialCategory}.`);
```
paulfouquet, line 101:
```suggestion
  throw new Error(`Path can't be generated from collection as no matching category for ${metadata.geospatialCategory}.`);
```

src/utils/date.ts
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * @param dateTimeString Optional date/time string which can be parsed by the `Date` constructor
+ * @returns Localised date/time string eg "2024-01-01"
+ *
+ */
+export function getPacificAucklandYearMonthDay(dateTimeString?: string | null): string | undefined {
+  if (dateTimeString == null) return;
+
+  // "sv-SE" formats date times as "yyyy-MM-dd hh:mm:ss"
+  const pacificAucklandDateTimeString = new Date(dateTimeString).toLocaleString('sv-SE', {
```
l0b0, line 12: Do we want to use [`toISOString`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString)? I don't really like depending on a specific locale, especially one which isn't local.

src/utils/__test__/date.test.ts
```diff
@@ -0,0 +1,11 @@
+import { describe, it } from 'node:test';
+
+import assert from 'assert';
+
+import { getPacificAucklandYearMonthDay } from '../date.js';
+
+describe('getPacificAucklandYearMonthDay', () => {
+  it('should format as yyyy-mm-dd', () => {
+    assert.equal(getPacificAucklandYearMonthDay('2012-12-31T11:00:00Z'), '2013-01-01');
```
l0b0, line 9: We probably want at least two more tests, to make sure time zones are handled properly:

- 2012-12-31T10:59:59Z (just before Pacific/Auckland midnight)
- 2012-06-15T11:59:59Z (just before Pacific/Auckland midnight in winter time)
- 2012-06-15T12:00:00Z (just at Pacific/Auckland midnight in winter time)

src/utils/date.ts
amfage: `getPacificAucklandYearMonthDay` is used by the [mapsheet-coverage command](https://github.com/linz/argo-tasks/blob/326adaf65094477536136b4cadb4a0cd2fcea8ca/src/commands/mapsheet-coverage/mapsheet.coverage.ts#L173). I've moved it in this PR because there was no need for date-related code in `generate-path` any more. @blacha could comment?

blacha: we cannot use `toISOString` as "The timezone is always UTC"

this formats the string as a almost ISO while putting the timezone as Pacific/Auckland.

l0b0 (dismissed):

paulfouquet (dismissed):

src/commands/generate-path/path.generate.ts
```diff
@@ -101,48 +77,39 @@ export const commandGeneratePath = command({
...
 
-  if (metadata.category === dataCategories.SATELLITE_IMAGERY) {
-    return `s3://${metadata.targetBucketName}/${metadata.region}/${surveyName}_${metadata.gsd}m/rgb/${metadata.epsg}/`;
+  if (
+    (
+      [
+        GeospatialDataCategories.UrbanAerialPhotos,
+        GeospatialDataCategories.RuralAerialPhotos,
+        GeospatialDataCategories.SatelliteImagery,
+      ] as string[]
```
blacha, line 91: I think the typing here is off ?

this shouldn't need a cast. Is something set as a "string" when it should be a `GeospatialDataCategories`

l0b0 (dismissed):

src/commands/generate-path/path.generate.ts
l0b0: Do you have a more specific suggestion? `as typeof GeospatialDataCategories[]` doesn't work:

> Argument type string is not assignable to parameter type {readonly SatelliteImagery: "satellite-imagery", readonly AerialPhotos: "aerial-photos", readonly RuralAerialPhotos: "rural-aerial-photos", readonly UrbanAerialPhotos: "urban-aerial-photos", readonly Dem: "dem", readonly ScannedAerialPhotos: "scanned-aerial-photos", readonly Dsm: "dsm"}

I also can't use `metadata.geospatialCategory as typeof GeospatialDataCategories`:

> Argument type {readonly SatelliteImagery: "satellite-imagery", readonly AerialPhotos: "aerial-photos", readonly RuralAerialPhotos: "rural-aerial-photos", readonly UrbanAerialPhotos: "urban-aerial-photos", readonly Dem: "dem", readonly ScannedAerialPhotos: "scanned-aerial-photos", readonly Dsm: "dsm"} is not assignable to parameter type "urban-aerial-photos" | "rural-aerial-photos" | "satellite-imagery"

blacha: I think it was mostly caused by type narrowing of the arrays, it is generally much more efficent to use `||` if statements than `.includes` They also fix the typing problem, 

I have pushed a commit to fix it.

blacha (dismissed):

blacha (approved):

PR #1152, opened 2024-12-02
feat: retile datasets TDE-1352

schmidtnz:
#### Motivation

As a user submitting datasets I want to poorly aligned source TIFF files to get aligned to our NZ 1:50k tile grid.

#### Modification

Updated `tileindex.validate.ts` to align input tiffs to our 1:50k tile grid.

#### Checklist

- [x] Tests updated
- [ ] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/src/commands/tileindex-validate/tileindex.validate.ts
+++ b/src/commands/tileindex-validate/tileindex.validate.ts
@@ -502,7 +502,10 @@ export function validateTiffAlignment(tiff: TiffLocation, allowedError = 0.015):
 export function getTileName(x: number, y: number, gridSize: GridSize): string {
   const sheetCode = MapSheet.sheetCode(x, y);
   // TODO: re-enable this check when validation logic
-  if (!MapSheet.isKnown(sheetCode)) throw new Error('Map sheet outside known range: ' + sheetCode);
+  if (!MapSheet.isKnown(sheetCode)) {
+    logger.warn('Map sheet outside known range: ' + sheetCode);
+    return '';
+  }
 
   // Shorter tile names for 1:50k
   if (gridSize === MapSheetTileGridSize) return sheetCode;
--- a/src/commands/tileindex-validate/tileindex.validate.ts
+++ b/src/commands/tileindex-validate/tileindex.validate.ts
@@ -1,5 +1,3 @@
-import assert from 'node:assert';
-
 import { Bounds, Projection } from '@basemaps/geo';
 import { fsa } from '@chunkd/fs';
 import { Size, Tiff, TiffTag } from '@cogeotiff/core';
@@ -225,19 +223,23 @@ export const commandTileIndexValidate = command({
     }
 
     const groupByTileNameStartTime = performance.now();
-    const locations = await extractTiffLocations(tiffs, args.scale, args.sourceEpsg);
+    const tiffLocations = await extractTiffLocations(tiffs, args.scale, args.sourceEpsg);
 
-    const outputs = groupByTileName(locations);
+    const outputTiles = groupByTileName(tiffLocations);
 
     logger.info(
-      { duration: performance.now() - groupByTileNameStartTime, files: locations.length, outputs: outputs.size },
+      {
+        duration: performance.now() - groupByTileNameStartTime,
+        files: tiffLocations.length,
+        outputs: outputTiles.size,
+      },
       'TileIndex: Manifest Assessed for Duplicates',
     );
 
     if (args.forceOutput || isArgo()) {
       await fsa.write('/tmp/tile-index-validate/input.geojson', {
         type: 'FeatureCollection',
-        features: locations.map((loc) => {
+        features: tiffLocations.map((loc) => {
           const epsg = args.sourceEpsg ?? loc.epsg;
           if (epsg == null) {
             logger.error({ source: loc.source }, 'TileIndex:Epsg:missing');
@@ -246,7 +248,6 @@ export const commandTileIndexValidate = command({
           return Projection.get(epsg).boundsToGeoJsonFeature(Bounds.fromBbox(loc.bbox), {
             source: loc.source,
             tileName: loc.tileNames.join(', '),
-            isDuplicate: (outputs.get(loc.tileNames.join(', '))?.length ?? 1) > 1,
           });
         }),
       });
@@ -254,11 +255,11 @@ export const commandTileIndexValidate = command({
 
       await fsa.write('/tmp/tile-index-validate/output.geojson', {
         type: 'FeatureCollection',
-        features: [...outputs.keys()].map((key) => {
+        features: [...outputTiles.keys()].map((key) => {
           const mapTileIndex = MapSheet.getMapTileIndex(key);
           if (mapTileIndex == null) throw new Error('Failed to extract tile information from: ' + key);
           return Projection.get(2193).boundsToGeoJsonFeature(Bounds.fromBbox(mapTileIndex.bbox), {
-            source: outputs.get(key)?.map((l) => l.source),
+            source: outputTiles.get(key)?.map((l) => l.source),
             tileName: key,
           });
         }),
@@ -267,38 +268,36 @@ export const commandTileIndexValidate = command({
 
       await fsa.write(
         '/tmp/tile-index-validate/file-list.json',
-        [...outputs.keys()].map((key) => {
-          const locs = outputs.get(key);
+        [...outputTiles.keys()].map((key) => {
+          const locs = outputTiles.get(key);
           return { output: key, input: locs?.map((l) => l.source) };
         }),
       );
-      logger.info({ path: '/tmp/tile-index-validate/file-list.json', count: outputs.size }, 'Write:FileList');
+      logger.info({ path: '/tmp/tile-index-validate/file-list.json', count: outputTiles.size }, 'Write:FileList');
     }
 
     let retileNeeded = false;
-    for (const val of outputs.values()) {
-      if (val.length < 2) continue;
-      // FIXME
+    for (const [tileName, tiffs] of outputTiles.entries()) {
+      if (tiffs.length < 2) continue;
       if (args.retile) {
-        const bandType = validateConsistentBands(val);
-        logger.info(
-          { tileName: val[0]?.tileNames[0], uris: val.map((v) => v.source), bands: bandType },
-          'TileIndex:Retile',
-        );
+        const bandType = validateConsistentBands(tiffs);
+        logger.info({ tileName, uris: tiffs.map((v) => v.source), bands: bandType }, 'TileIndex:Retile');
       } else {
         retileNeeded = true;
-        logger.error({ tileName: val[0]?.tileNames[0], uris: val.map((v) => v.source) }, 'TileIndex:Duplicate');
+        logger.error({ tileName, uris: tiffs.map((v) => v.source) }, 'TileIndex:Duplicate');
       }
     }
 
     // Validate that all tiffs align to tile grid
     if (args.validate) {
       let allValid = true;
-      for (const tiff of locations) {
-        const currentValid = validateTiffAlignment(tiff);
+      for (const tiffLocation of tiffLocations) {
+        const currentValid = validateTiffAlignment(tiffLocation);
         allValid = allValid && currentValid;
       }
-      if (!allValid) throw new Error(`Tile alignment validation failed`);
+      if (!allValid) {
+        throw new Error(`Tile alignment validation failed`);
+      }
     }
 
     if (retileNeeded) throw new Error(`Duplicate files found, see output.geojson`);
@@ -349,7 +348,7 @@ export interface TiffLocation {
   /** Location to the image */
   source: string;
   /** bbox, [minX, minY, maxX, maxY] */
-  bbox: [number, number, number, number];
+  bbox: BBox;
   /** EPSG code of the tiff if found */
   epsg?: number | null;
   /** Output tile name */
@@ -360,6 +359,43 @@ export interface TiffLocation {
    * @see {@link extractBandInformation} for more information on bad types
    */
   bands: string[];
+}
+
+/**
+ * Calculate the number grid tiles touched by a TIFF with min being the lowest width/height, and max the largest.
+ * @param inputMin lowest extent of misaligned tile on one dimension (left or bottom)
+ * @param inputMax largest extent of misaligned tile on same dimension (right or top)
+ * @param gridOrigin grid origin (x or y corresponding to min/max dimension)
+ * @param tileSize height or width of the target grid tile (corresponding to min/max dimension)
+ */
+const calculateSteps = (inputMin: number, inputMax: number, gridOrigin: number, tileSize: number): number => {
+  const startStep = Math.floor((inputMin - gridOrigin) / tileSize);
+  const endStep = Math.floor((inputMax - gridOrigin - 1) / tileSize);
+
+  return endStep - startStep + 1;
+};
+
+/**
+ * Reproject the bounding box if the source and target projections are different.
+ * @param bbox input bounding box
+ * @param sourceProjection CRS of the input bounding box
+ * @param targetProjection target CRS
+ */
+function reprojectIfNeeded(
+  bbox: BBox,
+  sourceProjection: Projection,
+  targetProjection: Projection,
+): [number | undefined, number | undefined, number | undefined, number | undefined] {
+  {
+    if (targetProjection !== sourceProjection) {
+      const [ulX, ulY] = targetProjection.fromWgs84(sourceProjection.toWgs84([bbox[0], bbox[3]]));
+      const [lrX, lrY] = targetProjection.fromWgs84(sourceProjection.toWgs84([bbox[2], bbox[1]]));
+
+      return [ulX, lrY, lrX, ulY];
+    }
+
+    return bbox;
+  }
 }
 
 /**
@@ -386,26 +422,10 @@ export async function extractTiffLocations(
           return null;
         }
 
-        const centerX = (bbox[0] + bbox[2]) / 2;
-        const centerY = (bbox[1] + bbox[3]) / 2;
-        // bbox is not epsg:2193
         const targetProjection = Projection.get(2193);
         const sourceProjection = Projection.get(sourceEpsg);
 
-        const [x, y] = targetProjection.fromWgs84(sourceProjection.toWgs84([centerX, centerY])).map(Math.round);
-        if (x == null || y == null) {
-          logger.error(
-            { reason: 'Failed to reproject point', source: tiff.source },
-            'Reprojection:ExtracTiffLocations:Failed',
-          );
-          return null;
-        }
-
-        // Tilename from center
-        const tileName = getTileName(x, y, gridSize);
-
-        const [ulX, ulY] = targetProjection.fromWgs84(sourceProjection.toWgs84([bbox[0], bbox[3]])).map(Math.round);
-        const [lrX, lrY] = targetProjection.fromWgs84(sourceProjection.toWgs84([bbox[2], bbox[1]])).map(Math.round);
+        const [ulX, lrY, lrX, ulY] = reprojectIfNeeded(bbox, sourceProjection, targetProjection);
 
         if (ulX == null || ulY == null || lrX == null || lrY == null) {
           logger.error(
@@ -416,7 +436,6 @@ export async function extractTiffLocations(
         }
 
         const covering = [...iterateMapSheets([ulX, ulY, lrX, lrY], gridSize)] as string[];
-        assert.ok(covering.includes(tileName));
 
         // if (shouldValidate) {
         //   // Is the tiff bounding box the same as the map sheet bounding box!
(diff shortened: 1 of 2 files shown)


src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -242,32 +243,31 @@ export const commandTileIndexValidate = command({
           }
           return Projection.get(epsg).boundsToGeoJsonFeature(Bounds.fromBbox(loc.bbox), {
             source: loc.source,
-            tileName: loc.tileName,
-            isDuplicate: (outputs.get(loc.tileName)?.length ?? 1) > 1,
+            tileName: loc.tileNames.join(', '),
```
schmidtnz, line 246: @blacha I'm not 100% sure if this is a sufficient solution with what that is being used for, or if the `tileName` attribute and `isDuplicate` flag need more attention?

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -276,15 +276,16 @@ export const commandTileIndexValidate = command({
     let retileNeeded = false;
     for (const val of outputs.values()) {
       if (val.length < 2) continue;
+      // FIXME
```
schmidtnz, line 279: @blacha what was the FIXME relating to? Does this code need further updates?

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -434,10 +471,13 @@ export function getSize(extent: [number, number, number, number]): Size {
 }
 
 export function validateTiffAlignment(tiff: TiffLocation, allowedError = 0.015): boolean {
-  const mapTileIndex = MapSheet.getMapTileIndex(tiff.tileName);
+  const tileName = tiff.tileNames[0];
+  // FIXME
```
schmidtnz, line 475: @blacha what was the FIXME relating to? Does this code need further updates?

blacha: I don't think we can assume that there is only one tileName since there is a array passed in.

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -522,3 +560,38 @@ export async function validate8BitsTiff(tiff: Tiff): Promise<void> {
...
+  const stepsY = calculateStep(minY, maxY, MapSheet.origin.y, tileHeight);
+
+  for (let stepY = 0; stepY < stepsY; stepY += 1) {
+    const y = maxY - stepY * tileHeight;
+    for (let stepX = 0; stepX < stepsX; stepX += 1) {
+      const x = minX + stepX * tileWidth;
+      try {
+        tile = getTileName(x, y, gridSize);
+      } catch (e) {
+        if (e instanceof RangeError) {
```
blacha, line 586: `instanceof` and Errors is somewhat risky pattern in javascript, in this particualr use case it would work, but its not a pattern we should be using.

I think we should just get rid of the is known check, we could turn it into a log and see how often its hit.

src/commands/tileindex-validate/tileindex.validate.ts
blacha: Yeah this would need a rethink when there are more than one output tiles, I think mostly just the isDuplicate, I dont think there is a good/easy way of representing arrays in geojson that shows up nicely in say qgis

src/commands/tileindex-validate/tileindex.validate.ts
blacha: we are assuming the tileNames is only one long, we should check

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -522,3 +560,38 @@ export async function validate8BitsTiff(tiff: Tiff): Promise<void> {
...
+          logger.warn(e['message']);
+          continue;
+        } else throw e;
+      }
+      yield tile;
+    }
+  }
+  if (tile === '') {
```
blacha, line 594: do we care if the iterator found no mapsheets? in the way we are using it, it would be cleaner to convert from a iterator into returning a full covering array.

src/commands/tileindex-validate/tileindex.validate.ts
schmidtnz: removed isDuplicate as discussed (this information can be gathered from other output files)

src/commands/tileindex-validate/tileindex.validate.ts
schmidtnz: now returning false early if length is not equal to 1.

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -478,14 +518,10 @@ export function validateTiffAlignment(tiff: TiffLocation, allowedError = 0.015):
...
-  const offsetY = Math.round(Math.floor((MapSheet.origin.y - y) / MapSheet.height));
-
-  // Build name
-  const letters = Object.keys(SheetRanges)[offsetY];
-  const sheetCode = `${letters}${`${offsetX}`.padStart(2, '0')}`;
-  // TODO: re-enable this check when validation logic
-  // if (!MapSheet.isKnown(sheetCode)) throw new Error('Map sheet outside known range: ' + sheetCode);
+  const sheetCode = MapSheet.sheetCode(x, y);
+  if (!MapSheet.isKnown(sheetCode)) {
+    logger.info(`Map sheet (${sheetCode}) at coordinates (${x}, ${y}) is outside the known range.`);
```
blacha, line 523: these should be extracted into log parameters, `logger.info({ sheetCode}, '...')`

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -222,80 +223,81 @@ export const commandTileIndexValidate = command({
...
       );
-      logger.info({ path: '/tmp/tile-index-validate/file-list.json', count: outputs.size }, 'Write:FileList');
+      logger.info({ path: '/tmp/tile-index-validate/file-list.json', count: outputTiles.size }, 'Write:FileList');
     }
 
     let retileNeeded = false;
-    for (const val of outputs.values()) {
-      if (val.length < 2) continue;
+    for (const [tileName, tiffs] of outputTiles.entries()) {
+      if (tiffs.length < 2) continue;
```
blacha, line 281: did wen want to make this `tiffs.length === 1` I think the tiffs.length === 0 is an assertion failure? so it should never happen.

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -357,6 +361,43 @@ export interface TiffLocation {
...
+  bbox: BBox,
+  sourceProjection: Projection,
+  targetProjection: Projection,
+): [number | undefined, number | undefined, number | undefined, number | undefined] {
+  {
+    if (targetProjection !== sourceProjection) {
+      const [ulX, ulY] = targetProjection.fromWgs84(sourceProjection.toWgs84([bbox[0], bbox[3]]));
+      const [lrX, lrY] = targetProjection.fromWgs84(sourceProjection.toWgs84([bbox[2], bbox[1]]));
+
+      return [ulX, lrY, lrX, ulY];
```
blacha, line 394: is this `[minX, minY, maxX, maxY]` we generally do Math.min/Math.max here.

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -381,23 +422,20 @@ export async function extractTiffLocations(
...
           logger.error(
             { reason: 'Failed to reproject point', source: tiff.source },
             'Reprojection:ExtracTiffLocations:Failed',
           );
           return null;
         }
 
-        // Tilename from center
-        const tileName = getTileName(x, y, gridSize);
+        const covering = [...iterateMapSheets([ulX, ulY, lrX, lrY], gridSize)] as string[];
```
blacha, line 438: would be good to just return a full `string[]` rather than iterating.

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -381,23 +422,20 @@ export async function extractTiffLocations(
...
 
-        const centerX = (bbox[0] + bbox[2]) / 2;
-        const centerY = (bbox[1] + bbox[3]) / 2;
-        // bbox is not epsg:2193
         const targetProjection = Projection.get(2193);
         const sourceProjection = Projection.get(sourceEpsg);
 
-        const [x, y] = targetProjection.fromWgs84(sourceProjection.toWgs84([centerX, centerY]));
-        if (x == null || y == null) {
+        const [ulX, lrY, lrX, ulY] = reprojectIfNeeded(bbox, sourceProjection, targetProjection);
```
blacha, line 428: we could stop destructuring this and leave it as a bbox (or eventually a polygon :) )

src/commands/tileindex-validate/tileindex.validate.ts
schmidtnz: Yes and no - I guess by doing this here without Math.min/Math.max we keep the order of the input intact, rather than taking it and possibly re-shuffling it by doing min/max.
It feels a tad safer than potentially re-ordering the data, especially as the other branch of the if statement does returns unchanged input bbox.

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -522,3 +560,46 @@ export async function validate8BitsTiff(tiff: Tiff): Promise<void> {
...
+  const stepsY = calculateSteps(-maxY, -bounds.y, -MapSheet.origin.y, tileHeight); // Inverted Y axis
+
+  for (let stepY = 0; stepY < stepsY; stepY += 1) {
+    const y = maxY - stepY * tileHeight;
+    for (let stepX = 0; stepX < stepsX; stepX += 1) {
+      const x = bounds.x + stepX * tileWidth;
+      const tileName = getTileName(x, y, gridSize);
+      const tile = MapSheet.getMapTileIndex(tileName);
+      if (tile == null) {
+        logger.error({ tileName, x, y, gridSize }, `No MapTileIndex for tile ${tileName} at xy ${x}, ${y}`);
```
blacha, line 585: does this ever happen? I feel like it should be a error?

src/commands/tileindex-validate/tileindex.validate.ts
schmidtnz: out of curiosity, what was the intent behind implementing this as an iterator at the start?

blacha: I believe I copy and pasted this function from somewhere it was doing a async process as it iterated, it may have also broken early rather than collecting the full list, but my memory is a little hazy as to exactly where it came from.

src/commands/tileindex-validate/tileindex.validate.ts
blacha: after reprojecting the top left may no longer be the minX, maxY, if we are being consistent that a bbox is minX, minY, maxX, maxY we need to confirm it.

src/commands/tileindex-validate/tileindex.validate.ts
schmidtnz: fixed in current code

src/commands/tileindex-validate/tileindex.validate.ts
schmidtnz: fixed

src/commands/tileindex-validate/tileindex.validate.ts
schmidtnz: adds another level of nesting .. but OK

blacha: another level of nesting?

```
if (tiffs.length === 0) throw new Error('output tile with no source tile: ...')
if (tiffs.length === 1) continue;

// ... mutilple input tiffs for a single output tiff
```

schmidtnz: my bad .. adding a level of nesting because I added a bug at the same time 👎🏿 
```
      if (tiffs.length === 1) {
        if (args.retile) {
```

schmidtnz: fixed

src/commands/tileindex-validate/tileindex.validate.ts
schmidtnz: fixed

src/commands/tileindex-validate/tileindex.validate.ts
schmidtnz: It can't really happen, as we are getting the `tileName` and feeding it straight into `getMapTileIndex`. 
But typescript wants me to check...
I have an idea how to implement (new MapSheet method) but probably outside the scope of this change.
Might make it an error.

schmidtnz: done

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -357,6 +362,35 @@ export interface TiffLocation {
   bands: string[];
 }
 
+/**
```
paulfouquet, line 365: An example and/or unit test might be helpful

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -522,3 +556,52 @@ export async function validate8BitsTiff(tiff: Tiff): Promise<void> {
     throw new Error(`${tiff.source.url.href} is not a 8 bits TIFF`);
   }
 }
+
+function getMapSheets(bbox: BBox, gridSize: GridSize): string[] {
```
paulfouquet, line 560: jsdoc comment, example/unit test would be helpful for this function

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -222,80 +223,82 @@ export const commandTileIndexValidate = command({
...
       );
-      logger.info({ path: '/tmp/tile-index-validate/file-list.json', count: outputs.size }, 'Write:FileList');
+      logger.info({ path: '/tmp/tile-index-validate/file-list.json', count: outputTiles.size }, 'Write:FileList');
     }
 
     let retileNeeded = false;
-    for (const val of outputs.values()) {
-      if (val.length < 2) continue;
+    for (const [tileName, tiffs] of outputTiles.entries()) {
+      if (tiffs.length === 0) throw new Error(`Output tile with no source tile: ${tileName}`);
```
blacha, line 281:
```suggestion
      if (tiffs.length === 0) throw new Error(`Output tile with no source tiff: ${tileName}`);
```

src/commands/tileindex-validate/tileindex.validate.ts
schmidtnz: done

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -357,6 +362,35 @@ export interface TiffLocation {
...
+ * @param bbox input bounding box
+ * @param sourceProjection CRS of the input bounding box
+ * @param targetProjection target CRS
+ */
+function reprojectIfNeeded(bbox: BBox, sourceProjection: Projection, targetProjection: Projection): BBox | null {
+  {
+    if (targetProjection !== sourceProjection) {
+      const [ulX, ulY] = targetProjection.fromWgs84(sourceProjection.toWgs84([bbox[0], bbox[3]]));
+      const [lrX, lrY] = targetProjection.fromWgs84(sourceProjection.toWgs84([bbox[2], bbox[1]]));
+      if (ulX === undefined || ulY === undefined || lrX === undefined || lrY === undefined) {
```
blacha, line 376:
```suggestion
      if (ulX == null || ulY == null || lrX == null || lrY == null) {
```

You often run into incorrect types where it says something is undefined, but in reality its null and vice versa.

src/commands/tileindex-validate/tileindex.validate.ts
schmidtnz: done

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -357,6 +362,35 @@ export interface TiffLocation {
...
+        logger.error(
+          {
+            reason: 'Failed to reproject point',
+            sourceBbox: bbox,
+            sourceProjection: sourceProjection,
+            targetProjection: targetProjection,
```
blacha, line 382:
```suggestion
            sourceProjection,
            targetProjection,
```

src/commands/tileindex-validate/tileindex.validate.ts
schmidtnz: changed

src/commands/tileindex-validate/tileindex.validate.ts
schmidtnz: added unit tests

src/commands/tileindex-validate/tileindex.validate.ts
schmidtnz: resolved this by asserting that we get `[number,number]`

src/commands/tileindex-validate/tileindex.validate.ts
schmidtnz: not needed now

src/commands/tileindex-validate/tileindex.validate.ts
schmidtnz: add jsdoc comment - this is being tested implicitly by the following tests, do you think more is needed?
`describe('TiffFromMisalignedTiff', () => {`

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -357,6 +362,21 @@ export interface TiffLocation {
...
+/**
+ * Reproject the bounding box if the source and target projections are different.
+ * @param bbox input bounding box
+ * @param sourceProjection CRS of the input bounding box
+ * @param targetProjection target CRS
+ */
+export function reprojectIfNeeded(bbox: BBox, sourceProjection: Projection, targetProjection: Projection): BBox | null {
+  {
+    if (targetProjection === sourceProjection) return bbox;
+    const [ulX, ulY] = targetProjection.fromWgs84(sourceProjection.toWgs84([bbox[0], bbox[3]])) as [number, number]; // Typescript inferred this as number[] but should be [number, number] | [number, number, number]
```
blacha, line 374: This comment is a little off, the typing of the function specifies number[] it could be refined to [number, number] when [number, number] is passed in, so the casting is fine

schmidtnz: I guess that's what the comment was trying to say, I have refined it from 
`// Typescript inferred this as number[] but should be [number, number] | [number, number, number]`
to
`// Function is typed as number[] but could be refined to [number, number] | [number, number, number]. With 2 input args, it will always return [number, number]`

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -522,3 +542,59 @@ export async function validate8BitsTiff(tiff: Tiff): Promise<void> {
...
+ *
+ * @param bbox Bounding box of the area of interest (in EPSG:2193) to get the map sheets for (e.g. TIFF area).
+ * @param gridSize Grid size of the map sheets / tiles to get.
+ * @param minIntersectionMeters Minimum intersection area in meters (width or height) to include the map sheet.
+ */
+function getCovering(bbox: BBox, gridSize: GridSize, minIntersectionMeters = 0.15): string[] {
+  const SurroundingTiles = [
+    { x: 1, y: 0 },
+    { x: 0, y: -1 }, // inverted Y axis
+    // Future versions may want to explore tiles in all directions
```
blacha, line 557: might be safer to leave these in? is there any downside? I feel like we could check a few million intersections a second?

schmidtnz: done

blacha (approved):

src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -522,3 +544,58 @@ export async function validate8BitsTiff(tiff: Tiff): Promise<void> {
...
+    const intersection = targetBounds.intersection(nextBounds);
+    if (intersection == null) continue; // no intersection, target mapshet is outside bounds of source
+    // Check all the surrounding tiles
+    for (const pt of SurroundingTiles) todo.push(nextBounds.add({ x: pt.x * tileWidth, y: pt.y * tileHeight })); // intersection not null, so add all neighbours
+    // Add to output only if the intersection is above the minimum coverage
+    if (intersection.width < minIntersectionMeters || intersection.height < minIntersectionMeters) continue;
+    output.push(getTileName(nextX, nextY, gridSize));
+  }
+
+  output.sort();
```
paulfouquet, line 599: nit: coud return as `sort()` [does also return the sorted array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#return_value)?

schmidtnz: done

paulfouquet (approved):

PR #1153, opened 2024-12-02
fix: remove GSD unit from stac setup command inputs TDE-1339 TDE-1211

amfage:
#### Motivation

A change was made recently to accept the GSD in the format without a trailing `m` e.g. `0.3` rather than `0.3m`. However, the old format should still be accepted as this may accidentally be entered as a workflow parameter.

#### Modification

Small refactor to match refactor made to `generate-path` command when handling `GeospatialDataCategories`.
Add a function `formatGsd` to check the format of the GSD, remove a trailing `m` and log a warning.

#### Checklist

- [x] Tests updated
- [x] Docs updated
- [x] Issue linked in Title

### DIFF
--- a/src/commands/common.ts
+++ b/src/commands/common.ts
@@ -169,12 +169,12 @@ export const UrlFolder: Type<string, URL> = {
  */
 export const MeterAsString: Type<string, string> = {
   from(str) {
-    const gsd = str.endsWith('m') ? str.slice(0, -1) : str;
+    const meters = str.endsWith('m') ? str.slice(0, -1) : str;
 
-    if (isNaN(Number(gsd))) {
-      throw new Error(`Invalid GSD value: ${gsd}. GSD must be a number.`);
+    if (isNaN(Number(meters))) {
+      throw new Error(`Invalid value: ${meters}. must be a number.`);
     }
 
-    return Promise.resolve(gsd);
+    return Promise.resolve(meters);
   },
 };
--- a/src/commands/common.ts
+++ b/src/commands/common.ts
@@ -159,3 +159,22 @@ export const UrlFolder: Type<string, URL> = {
     return url;
   },
 };
+
+/**
+ * Remove a trailing 'm' from a input value and validate the input is a number.
+ *
+ * @param str input value
+ * @returns value without trailing 'm' (if it exists)
+ * @throws if input is not a valid number
+ */
+export const MeterAsString: Type<string, string> = {
+  from(str) {
+    const gsd = str.endsWith('m') ? str.slice(0, -1) : str;
+
+    if (isNaN(Number(gsd))) {
+      throw new Error(`Invalid GSD value: ${gsd}. GSD must be a number.`);
+    }
+
+    return Promise.resolve(gsd);
+  },
+};
--- a/src/commands/stac-setup/__test__/stac.setup.test.ts
+++ b/src/commands/stac-setup/__test__/stac.setup.test.ts
@@ -207,13 +207,10 @@ describe('checkGsd', () => {
   });
 
   it('Should throw error if GSD is not a number', async () => {
-    await assert.rejects(
-      async () => await MeterAsString.from('foo'),
-      Error('Invalid GSD value: foo. GSD must be a number.'),
-    );
+    await assert.rejects(async () => await MeterAsString.from('foo'), Error('Invalid value: foo. must be a number.'));
     await assert.rejects(
       async () => await MeterAsString.from('1.4deg'),
-      Error('Invalid GSD value: 1.4deg. GSD must be a number.'),
+      Error('Invalid value: 1.4deg. must be a number.'),
     );
   });
 });
--- a/src/commands/stac-setup/__test__/stac.setup.test.ts
+++ b/src/commands/stac-setup/__test__/stac.setup.test.ts
@@ -205,7 +205,6 @@ describe('formatGsd', () => {
   });
 
   it('Should return GSD with trailing m removed', async () => {
-    const gsd = '0.3m';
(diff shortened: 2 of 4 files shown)


src/commands/stac-setup/__test__/stac.setup.test.ts
```diff
@@ -197,3 +197,15 @@ describe('formatDate', () => {
...
+
+describe('formatGsd', () => {
+  it('Should return GSD unaltered', async () => {
+    const gsd = '0.3';
+    assert.equal(formatGsd(gsd), '0.3');
+  });
+
+  it('Should return GSD with trailing m removed', async () => {
+    const gsd = '0.3m';
+    assert.equal(formatGsd(gsd), '0.3');
```
blacha, line 209: NIT: not much point for the variable?

```suggestion
    assert.equal(formatGsd('0.3m'), '0.3');
```

src/commands/stac-setup/__test__/stac.setup.test.ts
```diff
@@ -197,3 +197,15 @@ describe('formatDate', () => {
     assert.equal(formatDate(startYear, endYear), '2023-2024');
   });
 });
+
+describe('formatGsd', () => {
+  it('Should return GSD unaltered', async () => {
+    const gsd = '0.3';
+    assert.equal(formatGsd(gsd), '0.3');
```
blacha, line 204: NIT:

```suggestion
    assert.equal(formatGsd('0.3'), '0.3');
```

src/commands/stac-setup/stac.setup.ts
```diff
@@ -146,6 +143,20 @@ export function slugFromMetadata(metadata: SlugMetadata): string {
...
 
+/**
+ * Remove a trailing 'm' from a GSD value and log warning if it is present
+ *
+ * @param gsd GSD value from command line input
+ * @returns GSD without trailing 'm'
+ */
+export function formatGsd(gsd: string): string {
+  if (gsd.endsWith('m')) {
+    logger.warn(`${gsd} supplied as GSD; future supported format will require numerical value only.`);
```
paulfouquet, line 154: Do we have a date/ticket/issue that track this planned change?

paulfouquet: I think it would be good to describe what is a valid GSD value in [the description of the input parameter](https://github.com/linz/argo-tasks/blob/60501f1ea527ef629a8f5d2ce72506f7feae3c74/src/commands/stac-setup/stac.setup.ts#L44).

src/commands/stac-setup/stac.setup.ts
```diff
@@ -102,7 +102,7 @@ export const commandStacSetup = command({
         region: args.region,
         geographicDescription: args.geographicDescription,
         date: date,
-        gsd: args.gsd,
+        gsd: formatGsd(args.gsd),
```
paulfouquet, line 105: Do we want to add more verification/check on the inputs? e.g. if `gsd` is passed with any random string, what happens to the slug?

src/commands/stac-setup/stac.setup.ts
paulfouquet: This ticket is closed/"Done". How are we going to remember
> Deprecation notice in the code to remove at the next major release

?

src/commands/common.ts
```diff
@@ -159,3 +159,22 @@ export const UrlFolder: Type<string, URL> = {
...
+ * @param str input value
+ * @returns value without trailing 'm' (if it exists)
+ * @throws if input is not a valid number
+ */
+export const MeterAsString: Type<string, string> = {
+  from(str) {
+    const gsd = str.endsWith('m') ? str.slice(0, -1) : str;
+
+    if (isNaN(Number(gsd))) {
+      throw new Error(`Invalid GSD value: ${gsd}. GSD must be a number.`);
```
paulfouquet, line 175:
```suggestion
      throw new Error(`Invalid meters value: ${meters}. Must be a number.`);
```

src/commands/common.ts
```diff
@@ -159,3 +159,22 @@ export const UrlFolder: Type<string, URL> = {
...
+/**
+ * Remove a trailing 'm' from a input value and validate the input is a number.
+ *
+ * @param str input value
+ * @returns value without trailing 'm' (if it exists)
+ * @throws if input is not a valid number
+ */
+export const MeterAsString: Type<string, string> = {
+  from(str) {
+    const gsd = str.endsWith('m') ? str.slice(0, -1) : str;
```
paulfouquet, line 172:
```suggestion
    const meters = str.endsWith('m') ? str.slice(0, -1) : str;
```

src/commands/common.ts
```diff
@@ -159,3 +159,22 @@ export const UrlFolder: Type<string, URL> = {
...
+
+    if (isNaN(Number(gsd))) {
```
paulfouquet, line 174:
```suggestion
    if (isNaN(Number(meters))) {
```

src/commands/common.ts
```diff
@@ -159,3 +159,22 @@ export const UrlFolder: Type<string, URL> = {
...
+      throw new Error(`Invalid GSD value: ${gsd}. GSD must be a number.`);
+    }
+
+    return Promise.resolve(gsd);
```
paulfouquet, line 178:
```suggestion
    return Promise.resolve(meters);
```

src/commands/common.ts
blacha: I believe the tests look for this exact string :sob:

paulfouquet: Oh! That was just to align with the function becoming more "generic" that only `gsd`

paulfouquet (approved):

PR #1156, opened 2024-12-09
fix: include zero byte tiffs when validating files TDE-1348

blacha:
#### Motivation

zero byte tiffs are currently just skipped when validating files

#### Modification

allow file readers to opt into listing zero byte files

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title

### DIFF
--- a/src/utils/chunk.ts
+++ b/src/utils/chunk.ts
@@ -67,11 +67,15 @@ export type FileFilter = {
 
   /**
    * Limit the number of the output files
+   *
+   * @default -1 - No limit
    */
   limit?: number;
 
   /**
    * Group files into this number of items group
+   *
+   * @default -1 - No limit
    */
   group?: number;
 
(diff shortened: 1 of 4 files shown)


src/utils/chunk.ts
```diff
@@ -61,10 +61,52 @@ export function splitPaths(paths: string[]): string[] {
...
+   * @example
+   *
+   * ```typescript
+   * if(file.size < sizeMin) continue
+   * ```
+   * @default 1
+   */
+  sizeMin?: number;
+};
+export async function getFiles(paths: string[], args: FileFilter = { sizeMin: 1 }): Promise<string[][]> {
```
paulfouquet, line 106: Why not setting all the default value here?

blacha: This code block would be only triggered if args is undefined. I need to remove `sizeMin` from it, it was only put there for some temporary testing.

blacha, line 106:
```suggestion
export async function getFiles(paths: string[], args: FileFilter = {}): Promise<string[][]> {
```

paulfouquet (approved):

PR #1198, opened 2025-02-13
feat(stac-setup): support historical survey names LI-3977

blacha:
#### Motivation

We are still receiving more historical imagery we need to ensure we can continue to process the historical imagery

#### Modification

Add survey id to the list of optional arguments for stac-setup so that slugs can include the historical survery eg

```
west-coast_otago_sn8321_1984-1985_0.375m
west-coast_sn8534_1985-1986_0.375m
west-coast_sn9641_1997_0.75m
```

#### Verification

Unit tests

### DIFF
--- a/src/commands/stac-setup/__test__/stac.setup.test.ts
+++ b/src/commands/stac-setup/__test__/stac.setup.test.ts
@@ -20,18 +20,23 @@ describe('stac-setup', () => {
     mem.files.clear();
   });
 
+  const BaseArgs = {
+    verbose: false,
+    config: undefined,
+    surveyId: undefined,
+  };
+
   it('should retrieve setup from collection', async () => {
     const baseArgs = {
+      ...BaseArgs,
       odrUrl: 'memory://collection.json',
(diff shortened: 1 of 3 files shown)


src/commands/stac-setup/__test__/stac.setup.test.ts
```diff
@@ -182,18 +202,35 @@ describe('GenerateSlugDemIgnoringDate', () => {
...
-    assert.equal(formatDate(startYear, endYear), '2023');
+    assert.equal(formatDate('2023', '2023'), '2023');
   });
   it('Should return date as two years', async () => {
-    const startYear = '2023';
-    const endYear = '2024';
-    assert.equal(formatDate(startYear, endYear), '2023-2024');
+    assert.equal(formatDate('2023', '2024'), '2023-2024');
+  });
+  it('Should only return a data if both are set', () => {
```
blacha, line 228:
```suggestion
  it('Should only return a date if both are set', () => {
```

PR #1201, opened 2025-03-11
feat: node 23 direct typescript

blacha:
#### Motivation

Node23 has introduced running typescript files directly without need for compliation, to speed up developer experience and reduce the burden of building then running scripts, this pull request aims to migrate the entire repo into running directly from the source typescript.

for instance instead of building then running tests off build/**/*.test.ts tests can be run directly with node `node --watch src/utils/__test__/band.test.ts`

instead of building then running the cli, the cli can be started directly 

```
node src/index.ts \ 
  stac-validate \ 
  https://nz-imagery.s3-ap-southeast-2.amazonaws.com/new-zealand/new-zealand_2020-2021_10m/rgb/2193/collection.json
```

#### Modification

- Applies a `jscodeshift` to migrate all `.js` imports to `.ts` https://github.com/blacha/codeshift?tab=readme-ov-file#srcts-import-add-ext-tsjs
- Applies a eslint rule `@typescript-eslint/consistent-type-imports` to convert type imports into `import type` imports ref: https://nodejs.org/api/typescript.html#importing-types-without-type-keyword
- Moves all runtimes to `node23`
- Converts scripts to run off typescript source rather than `build/**/*.js`

#### Verification

container being built, some commands run locally


Dockerfile
```diff
@@ -16,9 +16,9 @@ ENV NODE_ENV production
...
 ADD package.json package-lock.json /app/
 RUN npm install --omit=dev
-ADD build/src /app/
+ADD src /app/
 
 # Cache of copy of the STAC JSON schemas by triggering a validation run
-RUN node /app/index.js stac-validate https://nz-imagery.s3-ap-southeast-2.amazonaws.com/new-zealand/new-zealand_2020-2021_10m/rgb/2193/collection.json
+RUN node /app/index.ts stac-validate https://nz-imagery.s3-ap-southeast-2.amazonaws.com/new-zealand/new-zealand_2020-2021_10m/rgb/2193/collection.json
 
-ENTRYPOINT ["node", "/app/index.js"]
```
Wentao-Kuang, line 24: This changes will require to update all the topo-workflow entry point

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@linzjs/argo-tasks",
   "private": true,
-  "version": "4.12.0",
```
Wentao-Kuang, line 4: Why we downgrade to `4.11,0`?

blacha: werid, looks to have been a bad merge somewhere.

Dockerfile
blacha: I wonder if we could avoid specifying the entrypoints in topo-workflow and use the defaults :shrug:

src/commands/basemaps-github/make.cog.github.ts
```diff
@@ -1,41 +1,51 @@
...
 
 export const DefaultCategorySetting: Record<Category, CategorySetting> = {
-  [Category.Urban]: { minZoom: 14 },
-  [Category.Rural]: { minZoom: 13 },
-  [Category.Satellite]: { minZoom: 5 },
-  [Category.Elevation]: { minZoom: 9 },
-  [Category.Scanned]: { minZoom: 0, maxZoom: 32 },
-  [Category.Other]: {},
-  [Category.Event]: {},
+  'Urban Aerial Photos': { minZoom: 14 },
```
Wentao-Kuang, line 42:
```suggestion
  ['Urban Aerial Photos']: { minZoom: 14 },
```
Or why we can't do `Categories.Urban` now?

package.json
blacha: Have fixed this now.

src/commands/basemaps-github/make.cog.github.ts
blacha: I went back and foward on a couple options here, maybe Categories.Urban is better ? but then Categories.Other is "New Aerial Photos" so the key value pairs are somewhat weirdly defined ?


Maybe I should make the categories more consistent with their actual values, but then I could just use the strings?

Dockerfile
blacha: looks like entrypoints are not required, so I have removed them in this pull request https://github.com/linz/topo-workflows/pull/985

src/commands/basemaps-github/make.cog.github.ts
Wentao-Kuang: Can We add the missing [] as well.

src/commands/basemaps-github/make.cog.github.ts
```diff
@@ -296,19 +306,19 @@ export class MakeCogGithub {
   async prepareElevationTileSetConfig(layer: ConfigLayer, tileSet?: ConfigTileSetRaster): Promise<ConfigTileSetRaster> {
     if (tileSet == null) {
       // Prepare individual elevation tileset config
-      const targetLayer = { ...layer, Category: Category.Elevation, minZoom: 0, maxZoom: 32 };
+      const targetLayer = { ...layer, Category: 'Elevation', minZoom: 0, maxZoom: 32 };
```
blacha, line 309: @Wentao-Kuang I noticed this had the wrong name, it should be `category` with a lower case `c` anythoughts on the impact by fixing it?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:20-slim@sha256:cffed8cd39d6a380434e6d08116d188c53e70611175cd5ec7700f93f32a935a6
+FROM node:23-slim
```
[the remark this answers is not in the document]
blacha: Good catch, I wonder if we can lint rule it so we have to do digests!?

src/commands/basemaps-github/make.cog.github.ts
blacha: the `[]` were not needed so I removed them.

src/commands/basemaps-github/make.cog.github.ts
Wentao-Kuang: I think that should be lower case, please fix it. And lucky we don't have individual Elevation tiles so far.

Dockerfile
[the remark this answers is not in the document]
blacha: fixed now.

src/commands/basemaps-github/__test__/create-pr-handler.test.ts
```diff
@@ -62,8 +61,8 @@ await it('basemapsCreatePullRequest.handler should handle S3 target', async (t)
     target: targetUrlsString,
     repository: 'any-owner/any-repository',
     verbose: false,
-    category: Category.Satellite,
-    configType: ConfigType.Raster,
+    category: 'Satellite Imagery',
```
schmidtnz, line 64: @blacha here and next line. Import removed from lines 11/12 (Category, ConfigType)

blacha: `category` is typed as `optional(oneOf(Object.values(Categories))),` so it has to be one of the Categories, so if you for instance say `Something Else` it would be a type error

schmidtnz (approved):

blacha:
> Do we want to update [the package `main` property](https://github.com/linz/argo-tasks/pull/1201/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R5) in this PR?

As this package is never published `("private": true)` these values are not used, I wonder if we should remove them?

PR #1207, opened 2025-04-01
feat: use `derived_from` checksum to see if tile needs re-creating TDE-1449

schmidtnz:
#### Motivation

As a data maintainer, when updating a national DEM Hillshade, I would like to recreate the hillshade tiles only for the nationally combined DEM tiles that have changed from the previous update.

#### Modification

Added command `identifyUpdatedItems` to identify for which items in a collection.json file there are updated items in the respective source collections (via derived_from links).

#### Checklist

- [x] Tests updated
- [x] Docs updated
- [x] Issue linked in Title


src/commands/item-changes/item.changes.ts
```diff
@@ -0,0 +1,157 @@
...
+    'Get a list of STAC items from source datasets that have changed or been added to the source compared to the target collection, based on existing hashes in STAC.',
+  version: CliInfo.version,
+  args: {
+    verbose,
+    targetCollection: option({
+      type: optional(string),
+      long: 'target-collection',
+      description: 'Target collection.json file that needs to be updated',
+    }),
+    sourceCollections: restPositionals({
```
paulfouquet, line 35: I don't know if it's "safe" to leave to the user to determine the Collection paths. I feel that you could use the `derived_from` link to identify the Collection path?

schmidtnz: A `derived_from` link points to the _source_ item (which will be in `sourceCollections`).
The `targetCollection` is the output collection.json that contains items with derived_from links (pointing to the items from `sourceCollections`).

`targetCollection` is optional so we can handle new collection.json publication - otherwise we would be limited to updating existing collections and would need different workflows. (A new publication is just an update where everything has changed).
If `targetCollection` wasn't optional, then yes we could use its content to determine which source collections to check, but it would be jumping from targetCollection to "item in target collection" to "item in source collection" to "source collection(s)" and we would have to do that for every item to make sure we don't miss any sourceCollections - I don't know if we would win much over the current solution?

(check how this gets called from the hillshade workflow to get an idea of its use)

src/commands/item-changes/item.changes.ts
```diff
@@ -0,0 +1,157 @@
...
+
+      return;
+    }
+    const targetCollectionUrl = args.targetCollection;
+    const sourceCollectionUrls = splitPaths(args.sourceCollections);
+    if (sourceCollectionUrls.length === 0 || sourceCollectionUrls.some((str) => !str.endsWith('collection.json'))) {
+      logger.error('Source collections must each point to existing STAC collection.json file(s)');
+      return;
+    }
+    type ItemSourceRecord = Record<string, { href: string; checksum: string }[]>;
```
paulfouquet, line 57: NIT: It's not useful to specify the type of the variable in its name, i.e. ItemSource**Record**. Something like `ItemSourceChecksums` would make more sense to me.

src/commands/item-changes/item.changes.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    const existingTargetItems: ItemSourceRecord = {};
```
paulfouquet, line 59: The type name is ItemSource* but the variable is a "Target". Maybe the type should have another name more generic?

src/utils/chunk.ts
```diff
@@ -134,3 +134,23 @@ export async function getFiles(paths: string[], args: FileFilter = {}): Promise<
...
 }
+
+/**
+ * Combine a base path with a relative path, resolving them to a single path.
+ *
+ * @param basePath The base path (can be a URL or local file path).
+ * @param addPath The path to combine with the base path. If this is an absolute path, it will overwrite basePath.
+ * @returns The combined absolute path.
+ */
+export function combinePaths(basePath: string, addPath: string): string {
```
paulfouquet, line 145: `basePath` is a bit confusing to me, as I expect a path like `/path/to/my/directory/` rather than `/path/to/my/file.ext`

schmidtnz: It's designed to work either way

src/commands/item-changes/item.changes.ts
schmidtnz: I really struggled finding a good name for ItemSourceRecord.
It's list of all locations (`href`) and respective checksums for the sources of each "item" (tile or mapsheet might be more appropriate, but they're STAC items so I went with item) together in one object.
ItemSource did not seem right, as it's not the source of an item (as in, it's not the STAC item json)

Any suggestions or ideas would be appreciated.

src/commands/item-changes/item.changes.ts
schmidtnz: I really struggled finding a good name for ItemSourceRecord.

`ItemSourceChecksums` sounds better.

src/commands/item-changes/item.changes.ts
schmidtnz: `Item` is the link between the two.
One is items existing in the target, and the other is a list of checksums for each source of the item.
If you have a suggestion how to handle this better, let me know :-)

schmidtnz: I've changed the variable names to put more emphasis on the `Item` part.

src/utils/chunk.ts
schmidtnz: Updated docstring to be more specific.

src/commands/identify-updated-items/identify.updated.items.ts
```diff
@@ -0,0 +1,167 @@
...
+  links: LinzItemLink[];
+}
+interface LinzStacCollection extends StacCollection {
+  links: LinzItemLink[];
+}
+
+export const commandIdentifyUpdatedItems = command({
+  name: 'identify-updated-items',
+  description:
+    'Get a list of STAC items from source datasets that have changed or been added to the source compared to the optional target collection, based on existing hashes in linked STAC documents. Note: If a target collection has been provided, its items links must be resolvable. If no target is specified, all items will be considered updated/new',
```
paulfouquet, line 27:
```suggestion
    'Compares the "derived_from" link checksums stored in the Items from the "target-collection" dataset against the current checksum of the Items of the "source-collection" dataset. Returns a list of the source Items that have changed or been added. Note: If a target collection has been provided, its Items links must be resolvable. If no target is specified, all source Items will be added to the list.',
```

src/commands/identify-updated-items/identify.updated.items.ts
```diff
@@ -0,0 +1,167 @@
...
+  version: CliInfo.version,
+  args: {
+    verbose,
+    targetCollection: option({
+      type: optional(string),
+      long: 'target-collection',
+      description:
+        'Target collection.json file that needs to be updated. If not provided, all items will be considered updated/new.',
```
paulfouquet, line 35:
```suggestion
        'Target collection.json file of the dataset that derives from the source dataset. If not provided, all Items of the source dataset will be returned in the list.',
```

src/commands/identify-updated-items/identify.updated.items.ts
```diff
@@ -0,0 +1,167 @@
...
+    }),
+    sourceCollections: restPositionals({
+      type: string,
+      displayName: 'source-collections',
+      description: 'Location of the source collection.json files. Split by ";"',
```
paulfouquet, line 40:
```suggestion
      description: 'Location of the source collection.json files that are used to create the target dataset. Split by ";"',
```

src/commands/identify-updated-items/identify.updated.items.ts
```diff
@@ -0,0 +1,167 @@
...
+      logger.error('Source collections must each point to existing STAC collection.json file(s)');
+      throw new Error('--source-collections must point to existing STAC collection.json file(s)');
+    }
+    type ItemSourceChecksums = Record<string, { href: string; checksum: string }[]>;
+
+    const existingItemsAtTarget: ItemSourceChecksums = {};
+    const desiredItemsAtTarget: ItemSourceChecksums = {};
+    const itemsToProcess: ItemSourceChecksums = {};
+
+    const Q = new ConcurrentQueue(10);
```
paulfouquet, line 63: Have you find that being a "good" number? 
If not sure and needing more adjustment, it would be an idea to makes it parameterable.

src/commands/identify-updated-items/identify.updated.items.ts
```diff
@@ -0,0 +1,167 @@
...
+    );
+
+    const fileListPath = '/tmp/identify-updated-items/file-list.json';
+    await fsa.write(fileListPath, JSON.stringify(tilesToProcess));
+    logger.info(
+      {
+        existingItems: Object.keys(existingItemsAtTarget).length,
+        desiredItems: Object.keys(desiredItemsAtTarget).length,
+        itemsToProcess: Object.keys(itemsToProcess).length,
+        itemsToProcessList: Object.keys(itemsToProcess).join(','),
```
paulfouquet, line 161: I wonder if we should log this information as it may be big...

src/commands/identify-updated-items/identify.updated.items.ts
```diff
@@ -0,0 +1,167 @@
...
+        ]),
+      ),
+      true,
+    );
+
+    const fileListPath = '/tmp/identify-updated-items/file-list.json';
+    await fsa.write(fileListPath, JSON.stringify(tilesToProcess));
+    logger.info(
+      {
+        existingItems: Object.keys(existingItemsAtTarget).length,
```
paulfouquet, line 158: I wonder if we should log this information as it may be big...

src/commands/identify-updated-items/identify.updated.items.ts
schmidtnz: The longest item here will be the comma-separated list of output tiles (line 161).
Compared to the amount of logging some other processes do, I think this is a concise summary of what is going on. 
Can limit the list to a fixed number of items if needed?

src/commands/identify-updated-items/identify.updated.items.ts
schmidtnz: The longest item here will be the comma-separated list of output tiles (line 161).
Compared to the amount of logging some other processes do, I think this is a concise summary of what is going on. 
Can limit the list to a fixed number of items if needed?

src/commands/identify-updated-items/identify.updated.items.ts
schmidtnz: It's an arbitrary (and low) number based on anecdotal evidence and https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html. Keeping this low (hence slowing down processing) is low impact and setting it too high may have unintended side-effects (do we handle 503 errors well?), so I'd rather leave it set to a low rate

schmidtnz: It's an arbitrary (and low) number based on anecdotal evidence and https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html. Keeping this low (hence slowing down processing) is low impact and setting it too high may have unintended side-effects (do we handle 503 errors well?), so I'd rather leave it set to a low rate

paulfouquet (approved):

PR #1214, opened 2025-05-01
feat: simple delete command TDE-1483

paulfouquet:
#### Motivation

In the context of the "Archiving Workflow", the system needs to delete the files after being archived. For simplicity in the tasks orchestration, the command `delete` takes a manifest of files that has been generated for the `copy` command.

#### Modification

- Add a command `delete` that takes a manifest of files that has been generated for the `copy` command. A `--dry-run` flag can be passed to only log the files that are going to be deleted.
 
> **_NOTE:_**  This PR is really opened to any suggestion to improve the way the command would be triggered and how the files to delete would be passed. It is currently using the same manifest of files that has been generated for the `copy` command and this is not ideal. First of all, it is tagged for the `copy` `action`. It also contains the `target` paths that won't be used for the `delete` but that is a potential severe issue/bug in future maintenance of the code.

#### Checklist

- [x] Tests updated
- [x] Docs updated
- [x] Issue linked in Title


src/commands/delete/delete.ts
```diff
@@ -0,0 +1,76 @@
...
+    const startTime = performance.now();
+    const pool = new WorkerRpcPool<DeleteContract>(args.concurrency, new URL('./delete-worker.ts', import.meta.url));
+
+    const stats = { deleted: 0, skipped: 0 };
+    const chunks = [];
+
+    for (const manifestPath of args.manifest) {
+      const json = await fsa.readJson<ActionCopy>(manifestPath);
+      //FIXME: Do we want a specific manifest for delete?
+      if (json.action !== 'copy') throw new Error('Invalid action: ' + String(json.action) + ' from:' + manifestPath);
```
paulfouquet, line 46: This does not feel right. It could need either a re-generation of a manifest file for the `delete` "action" or a modification from the `copy` command?

src/commands/delete/delete-worker.ts
```diff
@@ -0,0 +1,68 @@
...
+            return;
+          }
+
+          if (args.dryRun) {
+            logger.info({ path: todo.source }, 'File:Delete:SkippedDryRun');
+            stats.skipped++;
+            return;
+          }
+
+          await fsa.delete(todo.source);
```
paulfouquet, line 45: `todo` also contains the `target` that should not be deleted!

src/commands/delete/delete-rpc.ts
```diff
@@ -0,0 +1,23 @@
+export type DeleteContract = {
+  delete(args: DeleteContractArgs): Promise<DeleteStats>;
+};
+
+export interface DeleteStats {
+  /** Number of files deleted */
+  deleted: number;
+  /** Number of files that have been skipped (generally not found or dry-run) */
+  skipped: number;
```
schmidtnz, line 9: `copy` also tracks bytes - might be worth adding here?

src/commands/delete/delete-worker.ts
```diff
@@ -0,0 +1,68 @@
...
+import { ConcurrentQueue } from '../../utils/concurrent.queue.ts';
+import { registerCli } from '../common.ts';
+import type { DeleteContract, DeleteContractArgs, DeleteStats } from './delete-rpc.ts';
+
+const Q = new ConcurrentQueue(10);
+
+/** Current log id */
+let currentId: string | null = null;
+
+export const worker = new WorkerRpc<DeleteContract>({
```
blacha, line 16: Worker threads are generally if your are hitting a single threaded CPU limit (which it does with hashing), 

deletes shouldn't need use all the CPU Cores on the machine, you could greatly simplify this by removing all the worker logic

src/commands/delete/delete-worker.ts
blacha, line 45: S3 has a command to delete multiple objects in a single HTTP request, if performance is a problem, this would be a much faster way of deleting objects, (Im unsure if its cheaper).

https://docs.aws.amazon.com/cli/latest/reference/s3api/delete-objects.html

src/commands/delete/delete.ts
```diff
@@ -0,0 +1,76 @@
...
+
+const ManifestEntry = z.object({
+  source: z.string(),
+  target: z.string(),
+});
+const CopyManifest = z.array(ManifestEntry);
+
+export const commandDelete = command({
+  name: 'delete',
+  description: 'Delete source files listed in a copy manifest (TEMPORARY)', //FIXME: should be another kind of manifest?
```
blacha, line 21: I think the terminology should change to just a "manifest" as in a document containing a list of files.

I the manifest could just be the list of files.

src/commands/delete/delete-worker.ts
schmidtnz: There is a charge per request but [DELETE and CANCEL requests are free](https://aws.amazon.com/s3/pricing/). 

No doubt it will be significantly faster (only need to spin up a single node --> and hence also cheaper) to fire off 1 delete request instead of many, so this would greatly benefit from running a single delete command for all objects.

src/commands/delete/delete.ts
blacha: that would also work pretty well, the copy could output a delete manifest?

src/commands/delete/delete-worker.ts
blacha: > The request can contain a list of up to 1,000 keys that you want to delete. 

Im not entirely sure how long a DeleteObjects of 1,000 would take time wise!

schmidtnz: faster than waiting for (1000/group_size) ~25-250 pods to spin up? :-)

src/commands/delete/delete.ts
schmidtnz: or in this instance, the archive command can output the delete manifest.

src/commands/delete/delete-worker.ts
blacha: The group size here should be in the range of 100GB+? with small files that should be millions of deletes per pod.

schmidtnz: What I think we both want to say is, the `delete` action (and associated workflow) should not blindly follow what has been implemented in other commands (e.g. copy) or workflows (e.g. hillshade workflow has a group size of 4).

In addition to that, IMO we should be maxing out the number of items deleted per request to minimize the number of requests (as each request will come with an overhead).

src/commands/delete/delete.ts
paulfouquet: > the copy could output a delete manifest?

This will ensure the file that are going to be deleted have actually been copied. Downside is that each `copy` pod will output a "delete" manifest file VS if we let the `create-manifest` command to output both "copy" and "delete" manifests, we could have multiple "copy" manifest (as needs parallelism) and a single "delete" manifest (`delete` should be able to handle larger list of files - may need to cut into 1.000 chunks for AWS request - as per other comment on this PR)

paulfouquet:
Closing as deleting files will be handle by the `copy` command.

PR #1216, opened 2025-05-08
wip: compress and delete source files TDE-1482

schmidtnz:
#### Motivation

_What does this change aim to achieve?_

#### Modification

_Why is this change being made? What implications or other considerations are there?_

#### Checklist

- [ ] Tests updated
- [ ] Docs updated
- [ ] Issue linked in Title


src/commands/copy/copy-worker.ts
```diff
@@ -13,21 +15,31 @@ import { registerCli } from '../common.ts';
 import { isTiff } from '../tileindex-validate/tileindex.validate.ts';
 import type { CopyContract, CopyContractArgs, CopyStats } from './copy-rpc.ts';
 
-const Q = new ConcurrentQueue(10);
+const numberOfConcurrentItems = Math.max(cpus().length - 2, 1);
```
blacha, line 18: this script is invoked upto `concurrency` times per machine so max cpus is not a good metric to use especially when we run on machines with 100s of cores.

src/commands/copy/copy-worker.ts
```diff
@@ -1,5 +1,7 @@
+import { cpus } from 'node:os';
 import { performance } from 'node:perf_hooks';
 import { parentPort, threadId } from 'node:worker_threads';
+import zlib from 'node:zlib';
```
blacha, line 4: its best to import the specific items you are wanting with `{}`

src/commands/copy/copy-worker.ts
```diff
@@ -13,21 +15,31 @@ import { registerCli } from '../common.ts';
...
- * @returns New fixed file metadata if fixed other wise source file metadata
+ * @returns New fixed file metadata if fixed otherwise source file metadata
  */
 export function fixFileMetadata(path: string, meta: FileInfo): FileInfo {
-  // If the content is encoded we do not know what the content-type should be
-  if (meta.contentEncoding != null) return meta;
+  // Content-type is independent of content-encoding, so we should set both if possible. https://www.rfc-editor.org/rfc/rfc2616#section-14.11
+  if (path.endsWith('.zst')) {
+    // add appropriate encoding if file uses zstandard file extension
+    meta = { ...meta, contentEncoding: 'zstd' };
```
blacha, line 39: `contentEncoding` will mean this file is decompressed when accessed via a web browser do we really want that?

src/commands/copy/copy.ts
```diff
@@ -48,6 +48,21 @@ export const commandCopy = command({
...
+      defaultValueIsSerializable: true,
+    }),
+    deleteSource: flag({
+      type: boolean,
+      defaultValue: () => false,
+      long: 'delete-source',
+      description: 'Delete source files after successful copy or compress action',
+      defaultValueIsSerializable: true,
+    }),
     concurrency: option({ type: number, long: 'concurrency', defaultValue: () => 4 }),
```
blacha, line 66: I wonder if this should be `workers` or `threads` as the actual concurrency is `councurrency * workerConcurrency`

src/commands/copy/copy-worker.ts
```diff
@@ -13,21 +15,31 @@ import { registerCli } from '../common.ts';
 import { isTiff } from '../tileindex-validate/tileindex.validate.ts';
 import type { CopyContract, CopyContractArgs, CopyStats } from './copy-rpc.ts';
 
-const Q = new ConcurrentQueue(10);
+const numberOfConcurrentItems = Math.max(cpus().length - 2, 1);
+const Q = new ConcurrentQueue(numberOfConcurrentItems); // saturate all CPUs and leave 2 cores free for the system
+export const MIN_SIZE_FOR_COMPRESSION = 500; // testing with random ASCII data shows that compression is not worth it below this size
```
blacha, line 20: in this repo consts are `TitleCase` I wonder why the linter is not picking this up.

src/commands/copy/copy-worker.ts
```diff
@@ -99,42 +126,110 @@ export const worker = new WorkerRpc<CopyContract>({
...
+              ...manifestEntry,
+              size: source.size,
+              compressedSize: expectedSize,
+              ratio: ((expectedSize / source.size) * 100).toFixed(1) + '%',
+              duration: performance.now() - startTime,
+            },
+            'File:Compress:Done',
+          );
+        } else {
+          stats.copied++;
```
paulfouquet, line 215: Does that mean you would have to do `stats.copied + stats.compressed` to know the number of files that actually moved from source to target?

src/commands/copy/copy-worker.ts
amfage: Why is `TitleCase` used for exported consts in this repo rather than `SCREAMING_SNAKE_CASE`?

blacha: There is no real reason other than that is what is done in this repo, you can check other files searching for `export const` to see what things are exported as.

amfage: I just wondered why, but it's not important :)

src/commands/copy/copy-worker.ts
amfage: There is an Argo Workflows environment variable `AWS_TEMPLATE` that contains the number of CPUs requested for the pod container. We could use that to base the concurrency off rather than the CPU count (which would be the instance CPUs)?

src/commands/copy/copy-worker.ts
amfage: Unsure if this desired behaviour; will check how the files will be downloaded.

src/commands/copy/copy-worker.ts
amfage: I have made a change to combine the number of files uncompressed and compressed into `stats.copied`.

amfage: I think we should also change the following:
`stats.copiedBytes` (currently represents uncompressed bytes copied)
`stats.inputBytes` (bytes in the pre-compression input file)
`stats.outputBytes` (bytes in the post-compression output file)

src/commands/copy/copy.ts
amfage: Have updated command description/documentation.

src/commands/copy/copy-worker.ts
amfage: It seems it is better not to set the `Content-Encoding` in case the files are downloaded via the web interface (they may not be and in that case it would not make any difference anyway).

src/commands/copy/copy-worker.ts
schmidtnz: I've reverted this to a fixed queue size. Will open a new ticket for optimizing resource use.

src/commands/copy/copy-worker.ts
schmidtnz: @paulfouquet yes that was the intent. I found it would be useful to have the numbers split out so it is transparent what was done. This doesn't have an impact on the copy workflow, and in the archive workflow it would make it clear how much data was compressed vs copied uncompressed. The totals are probably less interesting than the individual numbers.

src/commands/copy/copy-worker.ts
schmidtnz: if we don't set the content encoding, we need to set the content type correctly (`octet-stream`).

src/commands/copy/copy-worker.ts
paulfouquet: I thought it would be useful to know how many files in total have been archived/moved (whether or not they have been compressed)

paulfouquet (approved):

blacha (approved):

PR #1221, opened 2025-05-19
fix: allow reading from complex 🦄🌈 URLs TDE-1504

blacha:
#### Motivation

Some input locations have very complex URLs, for example `s3://blacha-test/🦄 🌈/input.tiff`, when converting this location to a URL the path is URL encoded with `%`, when requesting URLs via the AWS SDK, the AWS SDK will automatically encode the URL again, causing a duplication of encoding. For example `%20` turns into `%2520` then S3 will fail to return the right object.

#### Modification

Add unit tests to validate our input and outputs to ensure URLS are not encoded for input and output
Add unit tests to validate commands can handle utf8 inputs 🦄🌈🦄🌈

#### Verification

ran unit tests + ran CLI locally

### DIFF
--- a/src/commands/tileindex-validate/__test__/tileindex.validate.test.ts
+++ b/src/commands/tileindex-validate/__test__/tileindex.validate.test.ts
@@ -12,7 +12,7 @@ import { MapSheetData } from '../../../utils/__test__/mapsheet.data.ts';
 import type { FileListEntry } from '../../../utils/filelist.ts';
 import type { GridSize } from '../../../utils/mapsheet.ts';
 import { MapSheet } from '../../../utils/mapsheet.ts';
-import { createTiff } from '../../common.ts';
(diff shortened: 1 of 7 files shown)


src/commands/tileindex-validate/__test__/tileindex.validate.test.ts
```diff
@@ -140,11 +140,12 @@ describe('tiffLocation', () => {
   });
 });
 
-describe('validate', () => {
+describe.only('validate', () => {
```
paulfouquet, line 143: Do not forget to remove `.only()` on the final version of you PR!

blacha, line 143:
```suggestion
describe('validate', () => {
```

paulfouquet (approved):
LGTM

PR #1226, opened 2025-05-22
fix: package information is not logged when command has not started TDE-1511

paulfouquet:
#### Motivation

If a command fails to start (for example, missing arguments), the package information (for example, `"package":{"package":"@linzjs/argo-tasks","version":"v4.14.2","hash":"e8dbf80c51b7a2baf61b22eec49de7072c812223","buildId":""}`) is not logged. Without this information, it makes it difficult to reproduce.

#### Modification

- Move package info logs before the command run

#### Verification

Local test

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,7 +8,7 @@ import { cmd } from './commands/index.ts';
 import { logger } from './log.ts';
 
 const startTime = performance.now();
-logger.info({ package: CliInfo }, 'Cli:Info');
+logger.info({ package: CliInfo, args: process.argv.slice(2) }, 'Cli:Info');
 run(cmd, process.argv.slice(2))
   .then(() => {
     logger.debug({ duration: performance.now() - startTime }, 'Command:Done');
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,10 +3,12 @@ process.env['AWS_NODEJS_CONNECTION_REUSE_ENABLED'] = '1';
 
 import { run } from 'cmd-ts';
 
+import { CliInfo } from './cli.info.ts';
 import { cmd } from './commands/index.ts';
 import { logger } from './log.ts';
 
 const startTime = performance.now();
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -3,10 +3,12 @@ process.env['AWS_NODEJS_CONNECTION_REUSE_ENABLED'] = '1';
 
 import { run } from 'cmd-ts';
 
+import { CliInfo } from './cli.info.ts';
 import { cmd } from './commands/index.ts';
 import { logger } from './log.ts';
 
 const startTime = performance.now();
+logger.info({ package: CliInfo }, 'Cli:Info');
```
blacha, line 11: random thought, do we also want to log `process.argv.slice(2)` (aka all the arguments?)

paulfouquet: I committed a change.
The arguments would be logged twice (but not identical way) if the command actually starts:
```shell
[2025-05-22T03:56:28.494Z] INFO Cli:Info id=01JVV259SX2JKR7PYR116GJKGV package={ package=@linzjs/argo-tasks version=TEST } args={ 0=stac-setup 1=--gsd 2=0.3 3=--region 4=new-zealand 5=--geospatial-category 6=dem }
[2025-05-22T03:56:28.499Z] INFO Cli:Start id=01JVV259SX2JKR7PYR116GJKGV cli=stac-setup args={ verbose=false gsd=0.3 region=new-zealand geospatialCategory=dem output=file:///tmp/stac-setup/ } isArgo=false
```

I am ok with that. Could maybe rename `args` to something else like `userInputs`?

blacha: How about a separate args logged at trace level, although I'm not sure we adjust the logger to trace until after the run, unless we have some sort of `TRACE=1` env var.

Either way maybe Im over thinking it and its too much effort, what are your thoughts?

schmidtnz (approved):

PR #1235, opened 2025-07-08
feat(basemaps): Pass ODR pull request url into basemap-config PR. BM-1324

Wentao-Kuang:
### Motivation

We want to pass a pull request body for the basemaps-config pr with the link of ODR pull request. So that we can pass the pr link via body from argo.

### Modifications

- Pass the body as cmd parameter 
- Output pr link from `stac-github-import cmd`, so that we can pass that into basemaps `create-pr`

### DIFF
--- a/src/commands/basemaps-github/__test__/create-pr-handler.test.ts
+++ b/src/commands/basemaps-github/__test__/create-pr-handler.test.ts
@@ -66,6 +66,7 @@ await it('basemapsCreatePullRequest.handler should handle S3 target', async (t) 
     individual: false,
     vector: false,
     ticket: 'any ticket',
+    content: 'any-content',
   });
   assert.equal(result, undefined);
   assert.equal(createPullRequestMock.mock.callCount(), 1);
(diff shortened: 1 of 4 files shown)


Wentao-Kuang:
> @Wentao-Kuang - Took me a moment to put all the parts together. I just want to check my understanding.
> 
> So, in the future, we will insert a task [here](https://github.com/linz/topo-workflows/blob/d9be58fe58abd179e6d819e6ad8ad3c07a45df08/workflows/basemaps/vector-etl-shortbread.yaml#L319) (between `analyse` and `create-pull-request`) to create an ODR pull request for the upcoming **topography** repo. This work modifies that argo-task to write the created PR's URL to a `tmp` location. From there, we'll consume the URL as the new `body` param in the existing create-pull-request task.
> 
> Will there be another step that allows us to "dress-up" the PR body further? Similar to your [example](https://github.com/linz/imagery/pull/649#issue-3194370308). Or, will the body just be the URL on its own?

That's completely correct, I was thinking just dress up a simple body with the ODR pull request link inside Argo parameter, and pass it into create-pr body. So we got a link in the basemaps-config PR.

PR #1238, opened 2025-07-21
refactor: copy command logic flow TDE-1589

schmidtnz:
### Motivation

Refactor the `copy` command logic to more easily support a future "decompression" option in preparation for implementing the decompression of archived files.

### Modifications
Extracted reusable logic from `copy-worker.ts` into new `copy-helpers.ts`.
Simplified logic to determine how to copy each file and updating stats.
### Verification

`npm run test` passes


src/commands/copy/copy-helpers.ts
```diff
@@ -0,0 +1,92 @@
...
+  return meta;
+}
+export async function determineTargetFileOperation(
+  source: FileInfo,
+  initialTargetName: string,
+  args: CopyContractArgs,
+): Promise<TargetFileOperation> {
+  const shouldCompress = args.compress && source.size !== undefined && source.size > MinSizeForCompression;
+  // const shouldDecompress = args.decompress && target.endsWith('.zst'); // TODO: Implement decompression logic
+  // const finalTargetName = shouldDecompress ? initialTargetName.slice(0, -COMPRESSED_FILE_EXTENSION.length) : initialTargetName + (shouldCompress ? COMPRESSED_FILE_EXTENSION : '');
```
amfage, line 47:
```suggestion
```

paulfouquet:
It would be good that the PR title reflects what's the refactor is aiming for, "refactor: refactor ..." is broad.

src/commands/copy/copy-helpers.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+export async function determineTargetFileOperation(
+  source: FileInfo,
+  initialTargetName: string,
+  args: CopyContractArgs,
+): Promise<TargetFileOperation> {
+  const shouldCompress = args.compress && source.size !== undefined && source.size > MinSizeForCompression;
+  const finalTargetName = initialTargetName + (shouldCompress ? CompressedFileExtension : '');
+
+  const target = (await fsa.head(finalTargetName)) || { path: finalTargetName, size: 0 };
```
blacha, line 48: `||` shouldn't be used, use `??`

src/commands/copy/copy-helpers.ts
```diff
@@ -0,0 +1,90 @@
...
+  const defaultOperation = shouldCompress ? FileOperation.Compress : FileOperation.Copy;
+  const myTargetAction: TargetFileOperation = {
+    target: target,
+    fileOperation: FileOperation.Skip,
+    shouldDeleteSourceOnSuccess: args.deleteSource || args.compress, // || shouldDecompress, // || shouldCompress,
+  };
+
+  if (!source.metadata) source.metadata = {};
```
blacha, line 56: avoid type cohersion
```suggestion
  if (source.metadata != null) source.metadata = {};
```

src/utils/file.head.ts
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * @param filePath File to head
+ * @param retryCount number of times to retry
+ * @returns file size if it exists or null
+ */
+export async function tryHead(filePath: string, retryCount = 3): Promise<FileInfo | null> {
+  for (let i = 0; i < retryCount; i++) {
+    const ret = await fsa.head(filePath);
+    if (ret?.size) return ret;
+    await new Promise((r) => setTimeout(r, 250));
```
blacha, line 17: Adjusting the backoff based off the retry logic could make this finish faster

```suggestion
    await new Promise((r) => setTimeout(r, 50 + i * 100));
```

src/commands/copy/copy-worker.ts
```diff
@@ -79,145 +32,117 @@ export const worker = new WorkerRpc<CopyContract>({
...
+          const hashCompressed = new HashTransform('sha256');
+
+          const rawSourceStream = fsa.stream(manifestEntry.source);
+          let sourceStream = rawSourceStream;
+
+          const shouldCompress = fileOperation === FileOperation.Compress;
+          if (fileOperation === FileOperation.Copy) {
+            sourceStream = rawSourceStream.pipe(hashOriginal);
+          } else if (shouldCompress) {
+            const zstd = createZstdCompress();
```
blacha, line 102: can we set the compression level here, what is the default?

src/commands/copy/copy-worker.ts
```diff
@@ -79,145 +32,117 @@ export const worker = new WorkerRpc<CopyContract>({
...
+        const startTime = performance.now();
         const source = await fsa.head(manifestEntry.source);
 
-        if (source == null) return;
-        if (source.size == null) return;
-        if (source.metadata == null) {
-          source.metadata = {};
+        if (source == null || source.size == null || source.size === 0) {
+          logger.info({ path: manifestEntry.source }, 'File:Copy:SkippedEmpty');
+          statsUpdaters[FileOperation.Skip]({ sourceSize: 0 });
```
blacha, line 73:
```suggestion
        const sourceSize = source?.size ?? 0
        if (sourceSize === 0) {
          logger.info({ path: manifestEntry.source }, 'File:Copy:SkippedEmpty');
          statsUpdaters[FileOperation.Skip]({ sourceSize });
```

src/commands/copy/copy-worker.ts
```diff
@@ -79,145 +32,117 @@ export const worker = new WorkerRpc<CopyContract>({
       skipped: 0,
       skippedBytes: 0,
     };
-    const end = Math.min(args.start + args.size, args.manifest.length);
+    const statsUpdaters = {
```
blacha, line 35: if you type this as `Reacord<FileOperation, (sourceSize, outputSize) => void>` you could drop all the types, 

also do these need to be objects it adds a fair bit of overhead

blacha: also if this took a `stats` as a object, it makes it really easy to test

`(stats: Stats, sourceSize: number, outputSize:number) => void`

src/commands/copy/copy-helpers.ts
schmidtnz: Note - the suggestion above is inverting the logic, so not accepting that. 
Simplified to:
```suggestion
  source.metadata ??= {};
```

src/commands/copy/copy-helpers.ts
schmidtnz: `||` is used here to guarantee that at least `path` and `size` will be present, even if a valid but empty metadata object was returned `fsa.head`. 
Can you give an example how that can be achieved using `&&`?

schmidtnz: would this be a good approach?
```
const target = {
  path: finalTargetName,
  size: 0,
  ...(await fsa.head(finalTargetName)) ?? {},
};
```

src/commands/copy/copy-worker.ts
schmidtnz: This doesn't narrow the type of `source` - I'll submit a slightly modified version that is more concise than what I've currently got.

src/commands/copy/copy-worker.ts
schmidtnz: Yes, options can be passed in here.
Set compression parameters according to pre-defined cLevel table. Default level is ZSTD_CLEVEL_DEFAULT==3.

src/commands/copy/copy-helpers.ts
blacha: your quite correct the check should have been `== null`

src/commands/copy/copy-helpers.ts
blacha: `||` has flaws in its logic when used  with nullish assignment especially when working with numbers and generally should be avoided. `??` is the more modern apporach

stylistically I don't love `(await foo) ?? bar` 

I would generally write this as something more like

```
const head = await fsa.head(targetName);
const size = head?.size ?? 0
```

But I'm not too sure exactly what impacts that would have on the rest of this function

src/commands/copy/copy-worker.ts
blacha: We should revisit this number 3 is very low :) (in a future PR)

src/commands/copy/copy-helpers.ts
```diff
@@ -0,0 +1,90 @@
...
+  if (path.endsWith('.json')) return { ...meta, contentType: 'application/json' };
+
+  return meta;
+}
+export async function determineTargetFileOperation(
+  source: FileInfo,
+  initialTargetName: string,
+  args: CopyContractArgs,
+): Promise<TargetFileOperation> {
+  const shouldCompress = args.compress && source.size !== undefined && source.size > MinSizeForCompression;
```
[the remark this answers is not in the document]
schmidtnz: resolved by splitting it into a function and use of default values

src/commands/copy/copy-helpers.ts
schmidtnz: resolved without `(await foo) ?? bar` while retaining default values

src/commands/copy/copy-worker.ts
schmidtnz: are you going to add a ticket for this? We've already seen some pretty good compression ratios using default.

src/commands/copy/copy-worker.ts
schmidtnz: done

src/commands/copy/copy-rpc.ts
```diff
@@ -49,3 +59,22 @@ export interface CopyStats {
...
+export const FileOperation = {
+  Copy: 'copy',
+  Skip: 'skip',
+  Compress: 'compress',
+  Delete: 'delete', // pseudo-operation for deleting source files after other operations
+} as const;
+export type FileOperation = (typeof FileOperation)[keyof typeof FileOperation];
+
+export interface TargetFileOperation {
+  /**
```
paulfouquet, line 72: Is this doc meant to be here?

paulfouquet (approved):

amfage (approved):

PR #1239, opened 2025-07-24
feat: validate restored state of S3 objects from S3 Batch restore report TDE-1554

paulfouquet:
### Motivation

S3 Batch Operations generate restore reports, but these only confirm that the restore request was accepted — not that the objects were actually restored and available for reading. We need to ensure that every object listed in a successful restore report is truly restored, before continuing with downstream workflows (e.g. copying to a shared bucket).
This reduces the risk of trying copying unavailable or still-restoring objects.

### Modifications

Added a `verify-restore` command that:

- Reads a given S3 Batch Operations restore manifest (`manifest.json`)
- Parses the result CSV(s) listed in the manifest
- Verifies that all entries have `TaskExecutionStatus = succeeded`
- Checks that the objects are actually accessible and restored
- Throws an error if any requested objects are missing or the restoration status can't be accessed
- Outputs a the status of the restoration (`true` for completed, `false` for still on-going) in a file
- If `--mark-done` is passed and the verification successful, rename the report with the suffix `*.done` to prevent reprocessing

### Verification

- Unit tests
- Local tests
- Argo Workflows: TODO


src/commands/verify-restore/verify.restore.ts
```diff
@@ -0,0 +1,230 @@
...
+}
+
+/**
+ * Heads an S3 object.
+ *
+ * @param path - The S3 path to the object to check.
+ * @throws Will throw an error if the headObject request fails.
+ * @returns The head object output.
+ */
+async function headObject(path: { Bucket: string; Key: string }): Promise<HeadObjectCommandOutput> {
```
schmidtnz, line 191: Naming - might be good to be more explicit how it is different to `fsa.head()` (specific to s3 metadata).

schmidtnz (approved):

PR #1244, opened 2025-08-06
chore: update node.js to 24.5 to match gh actions TDE-1600

schmidtnz:
### Motivation

Our current github action builds are failing.

### Modifications

Updated package.json to the current node version used by github actions (24.5.0)
Updated package-lock.json
Pinned node versions in github actions to 24.5.0 (was 24.x)

### Verification

builds succeed.

### DIFF
--- a/.github/workflows/pull-request-container.yml
+++ b/.github/workflows/pull-request-container.yml
@@ -26,7 +26,7 @@ jobs:
     steps:
       - uses: linz/action-typescript@dee99184c4305aea6c380a52db9b2d7abaaa3e78 # v3
         with:
-          node-version: 24.x
+          node-version: 24.5.0
 
       - name: Setup GIT version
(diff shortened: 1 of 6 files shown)


.github/workflows/pull-request-container.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     steps:
       - uses: linz/action-typescript@dee99184c4305aea6c380a52db9b2d7abaaa3e78 # v3
         with:
-          node-version: 24.x
+          node-version: 24.5.0
```
amfage, line 29: Can we specify `24.5.x` so we get patch releases?

schmidtnz: With this being our prod environment, I'd prefer to pin to a specific version

PR #1256, opened 2025-09-05
feat: delete empty folders TDE-1627

schmidtnz:
### Motivation
When running an [archive workflow](https://github.com/linz/topo-workflows/tree/master/workflows/storage#archive) it can happen that some ["folder" S3 objects](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-folders.html) remain in the source location (and the source itself) of files being archived. This is due to manually created "folders" in S3 when original files have been uploaded. In most cases, we would like to remove these empty folders from the source location.

_Based on https://github.com/linz/argo-tasks/pull/1254 with updated chunkd._

### Modifications
Add a `delete-empty-folders` command that deletes empty folders and "folder" S3 objects

### Verification
Automated tests with local filesystem.
Manual tests with local and S3 locations.
Argo Workflows


src/commands/delete-empty-folders/delete.empty.folders.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  let allEmpty = true;
+  const deletedFolders: URL[] = [];
+
+  for (const obj of subs) {
+    if (await isFolder(obj)) {
+      if (obj.href === location.href) continue;
+      const subDeleted = await deleteEmptyFolders(obj, dryRun);
+      deletedFolders.push(...subDeleted);
+      // if (subDeleted.some((item) => item.href === obj.href)) {
```
amfage, line 64: Sneaky little comment here

schmidtnz: fixed -- it didn't want to be a comment :-)

amfage (approved):

PR #1265, opened 2025-09-16
fix: accept aerial-photos as category TDE-1650

schmidtnz:
### Motivation
aerial-photos is a category available in our argo workflows, but the standardising workflow fails to create STAC when selected.

### Modifications
Add `GeospatialDataCategories.AerialPhotos` as a valid category in function `generatePath` and appropriate test.

### Verification

Passing unit tests.

### DIFF
--- a/src/commands/generate-path/__test__/generate.path.test.ts
+++ b/src/commands/generate-path/__test__/generate.path.test.ts
@@ -24,6 +24,18 @@ describe('GeneratePathImagery', () => {
     };
     assert.equal(generatePath(metadata), 's3://nz-imagery/auckland/auckland_2023_0.3m/rgb/2193/');
   });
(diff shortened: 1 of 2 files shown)


src/commands/generate-path/__test__/generate.path.test.ts
```diff
@@ -24,6 +24,18 @@ describe('GeneratePathImagery', () => {
     };
     assert.equal(generatePath(metadata), 's3://nz-imagery/auckland/auckland_2023_0.3m/rgb/2193/');
   });
+  it('Should match - generic aerial from slug', () => {
```
amfage, line 27:
```suggestion
  it('Should match - generic aerial photos from slug', () => {
```

amfage (approved):

PR #1268, opened 2025-09-18
feat: tileindex-validate to output GSD TDE-1655

schmidtnz:
### Motivation

As an imagery maintainer
I want to make sure the resolution of all TIFF images matches
so that the data is consistent and accurate (in line with the metadata)

This change is to validate all TIFFs have a consistent GSD and make the GSD available to other workflow steps.

### Modifications

Extract GSD (resolution) from all tiff files that get processed.
Output GSD to a file for use by argo workflows

Added tests.

### Verification

Unit tests.

### DIFF
--- a/src/commands/tileindex-validate/tileindex.validate.ts
+++ b/src/commands/tileindex-validate/tileindex.validate.ts
@@ -227,11 +227,22 @@ export const commandTileIndexValidate = command({
     const tiffs = await TiffLoader.load(tiffLocationsToLoad, args);
     await validatePreset(args.preset, tiffs);
 
-    const projections = new Set(tiffs.map((t) => t.images[0]?.epsg));
+    const projections = new Set();
+    const gsds = new Set();
+    const roundedGsds = new Set();
+    tiffs.forEach((t) => {
+      const image = t.images[0];
+      if (image) {
+        projections.add(image.epsg);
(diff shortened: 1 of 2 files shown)


src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -240,6 +251,18 @@ export const commandTileIndexValidate = command({
...
       logger.warn({ projections: [...projections] }, 'TileIndex:InconsistentProjections');
     }
+    if (roundedGsds.size > 1) {
+      if (args.validate) {
+        logger.error({ gsds: [...gsds], roundedGsds: [...roundedGsds] }, 'TileIndex:InconsistentGSDs:Failed');
+        throw new Error(
+          `Inconsistent GSDs found: ${[...roundedGsds].join(', ')} ${[...gsds].join(',')}, ${tiffLocationsToLoad.map(protocolAwareString).join(',')}`,
+        );
+      }
+      logger.warn({ gsds: [...gsds], roundedGsds: [...roundedGsds] }, 'TileIndex:InconsistentGSDs:Failed');
```
amfage, line 261: Noting as discussed it would be good to alert on this or explore setting exit status so that users can be notified if there is a GSD mismatch.

PR #1269, opened 2025-09-19
feat: Stac setup to check GSD against collection.json TDE-1654

schmidtnz:
### Motivation
When provided an ODR URL, stac setup should optionally validate that the GSD received as an input matches the existing data in ODR (collection.json or failing that, existing TIFF files).

### Modifications
Added optional `validate` flag.
Read back GSD from collection.json or first TIFF in collection.

### Verification
added unit tests


src/commands/stac-setup/stac.setup.ts
```diff
@@ -59,6 +60,14 @@ export const commandStacSetup = command({
       description: 'GSD of dataset, e.g. 0.3',
     }),
 
+    validate: flag({
+      type: boolean,
+      defaultValue: () => false,
+      long: 'validate',
```
amfage, line 66: NIT it

amfage: Can be revisited in future if necessary.

src/commands/stac-setup/stac.setup.ts
```diff
@@ -59,6 +60,14 @@ export const commandStacSetup = command({
...
+      description: 'Validate GSD of dataset if odr_url is supplied',
```
paulfouquet, line 67: This is confusing as there is no `odr_url` to supply here?

paulfouquet: Oh that is `odr-url`. I did a dumb search.

schmidtnz: I noticed the inconsistent spelling as I pasted the link ...

PR #1271, opened 2025-09-22
feat: output resolution for resampling TDE-1658

schmidtnz:


schmidtnz:
implemented on the workflow side without additional changes in argo tasks

PR #1311, opened 2026-05-11
feat: allow 16 bit tiffs with rgb+nir

blacha:
### Motivation

In some cases we should be allowing 16bit rgbinir, to test the process temporarlly allow 16bit nir to pass the validation logic

### Modifications

allow rgbnir datasets to be either 8bit or 16bit provided all bands are consistent

### DIFF
--- a/src/commands/tileindex-validate/__test__/tileindex.validate.test.ts
+++ b/src/commands/tileindex-validate/__test__/tileindex.validate.test.ts
@@ -25,6 +25,7 @@ import {
   TiffLoader,
   validate8BitsTiff,
   validatePreset,
+  validateTiffSamples,
 } from '../tileindex.validate.ts';
 import { FakeCogTiff } from './tileindex.validate.data.ts';
 
@@ -263,6 +264,8 @@ describe('is8BitsTiff', () => {
       name: 'Error',
       message: `${process.cwd()}/src/commands/tileindex-validate/__test__/data/16b.tiff is not a 8 bits TIFF`,
     });
+
+    assert.ok(await validateTiffSamples(testTiff, new Set([16])));
   });
 });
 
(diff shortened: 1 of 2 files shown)


src/commands/tileindex-validate/tileindex.validate.ts
```diff
@@ -473,18 +479,34 @@ export function determineGridSizeFromGSDPreset(gsd: number, preset: string): Gri
...
+      rejected = true;
     });
-    await Promise.allSettled(promises);
-  }
+    if (value == null) return;
+    if (bitCount == null) bitCount = value;
+    if (bitCount !== value) {
+      logger.fatal(
+        {
+          reason: `${protocolAwareString(t.source.url)} Inconsistent bit depth across bands: ${bitCount} vs ${value}}`,
```
amfage:
```suggestion
          reason: `${protocolAwareString(t.source.url)} Inconsistent bit depth across bands: ${bitCount} vs ${value}`,
```

src/commands/tileindex-validate/__test__/tileindex.validate.test.ts
```diff
@@ -271,15 +275,15 @@ describe('validatePreset', () => {
     const test16bTiff = await createTiff(pathToFileURL('./src/commands/tileindex-validate/__test__/data/16b.tiff'));
     const test8bTiff = await createTiff(pathToFileURL('./src/commands/tileindex-validate/__test__/data/8b.tiff'));
     const fatalStub = t.mock.method(logger, 'fatal');
-    await assert.rejects(validatePreset(preset, [test16bTiff, test16bTiff, test8bTiff]), {
+    await assert.rejects(validatePreset(preset, [test8bTiff, test16bTiff, test16bTiff]), {
```
amfage, line 278: Isn't `[test16bTiff, test16bTiff, test8bTiff]` a scenario that could happen? Wouldn't this give a different result?

blacha: I would assume both of these would error as the bit count is inconsistent, My memory is a little hazy, is there a reason you want the test to be this way arround?

amfage: Actually I think it will be fine I was just thinking because of the asynchronous nature it could be reversed but I don't think it matters.

amfage (approved):

PR #1328, opened 2026-07-20
feat: Chatham Islands EPSG:3793 and mapsheets TDE-1990

amfage:
### Motivation

In order to standardise Chatham Islands datasets, EPSG 3793 (CITM2000) needs to be accepted and the relevant map sheet origins supported.

### Modifications

Look up table for Chatham Islands map sheets, adding support for EPSG 3793.

### Verification

Unit tests and workflow run.

### DIFF
--- a/src/utils/mapsheet.ts
+++ b/src/utils/mapsheet.ts
@@ -360,15 +360,25 @@ export interface ChathamSheetDefinition {
 }
 
 /**
- * Static definition of the six Chatham Islands Topo50 map sheets (EPSG:3793, NZGD2000 / Chatham
- * Islands TM 2000), laid out on the same 24,000m x 36,000m 1:50k grid as the mainland
- * {@link MapSheet}, just with a different origin, CRS, and (small, fixed) set of sheet codes.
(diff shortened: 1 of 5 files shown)


src/utils/mapsheet.ts
```diff
@@ -305,3 +363,129 @@ export const SheetRanges = {
...
+   * ChathamMapSheet.offset("CI06") // { x: 3506000, y: 5104000 }
+   * ```
+   * @throws if the sheet code is not one of the six known Chatham Islands sheets, see {@link ChathamMapSheet.isKnown}
+   */
+  offset(sheetCode: string): Point {
+    const sheet = ChathamSheetByCode.get(sheetCode.slice(0, 4));
+    if (sheet == null) {
+      throw new Error(`Unknown Chatham Islands map sheet "${sheetCode}"; not one of the six known sheets`);
+    }
+    return sheet.origin;
```
schmidtnz: nit - copy/return a new object to avoid mutation

```suggestion
    return { ...sheet.origin };
```

schmidtnz (approved):

